PR #116, opened 2015-01-14
Small refactoring the export mechanism

micheles:
Now the key used in the export dictionary is not a string of the form 'name_ext' but a pair ('name', 'ext').
This is done for compatibility with the engine.
The tests are green: https://ci.openquake.org/job/zdevel_oq-risklib/181/


ptormene:
LGTM

PR #118, opened 2015-01-18
Fixed the file fragility_continuous.xml

micheles:
Somebody very bad happened here. The file fragility_continuous.xml was incorrect, missing the attributes minIML and maxIML, however Jenkins did not raise an error (!??)
See for instance https://ci.openquake.org/job/master_oq-engine/1508/consoleFull, where the test that would break (ScenarioDamageTestCase.test_case_2) for some
reason did not run. But if you run the test on a machine different from Jenkins it breaks, as it should. This is very ugly and dangerous, since bad code entered in master
without anybody noticing.


micheles:
LGTM since this is not a code change

PR #119, opened 2015-01-18
Add a test case for sampling two source models of different weight

micheles:
This is preliminary work for the undersampling bug https://bugs.launchpad.net/oq-engine/+bug/1408313. It also fixes a serious bug in parallel.apply_reduce when concurrent_tasks=0. The tests are green, https://ci.openquake.org/job/zdevel_oq-risklib/184/ including the tests on the engine side https://ci.openquake.org/job/zdevel_oq-engine/964/


ptormene:
LGTM

PR #133, opened 2015-01-30
Raised the tolerance of the assert in scientific.mean_curve

micheles:
Before it was 1E-15, too low, and we got spurious errors due to rounding issues in weights which were actually were very close to 1.


mmpagani:
LGTM!!

PR #138, opened 2015-02-04
Towards supporting Ubuntu 14.04

micheles:
Here is a fix to cope with the API change in psutil. The tests are green: https://ci.openquake.org/job/zdevel_oq-risklib/236


daniviga:
LGTM if CI is green

PR #144, opened 2015-02-09
Fixed the test case_13 to work with newer libraries

micheles:
See https://bugs.launchpad.net/oq-engine/+bug/1418961. By removing the points with coordinates -121.57 37.06 and 121.76 37.23 we can hide the discrepancy under the carpet.
The problem is the exposed in a specific test, the one introduced here: https://github.com/gem/oq-hazardlib/pull/287


daniviga:
LGTM

PR #146, opened 2015-02-10
Classical tiling calculator in oq-lite

micheles:
See https://bugs.launchpad.net/oq-engine/+bug/1419398.
I have also added a check so that, if `mean_hazard_curves` is False, the mean curves are not computed nor exported.
The tests are running here: https://ci.openquake.org/job/zdevel_oq-risklib/254


micheles:
LGTM, since this is needed urgently to fix an error in master: https://ci.openquake.org/job/master_oq-engine/1550/

PR #152, opened 2015-02-16
Enhanced the baselib.general.groupby facility

micheles:
This is a small improvement.


ptormene:
If this doesn't break any test, LGTM

PR #155, opened 2015-02-18
Fix a bug with the weights in case of sampling

micheles:
See https://bugs.launchpad.net/oq-engine/+bug/1422631. The tests are running here: https://ci.openquake.org/job/zdevel_oq-risklib/270/

### DIFF
--- a/openquake/commonlib/source.py
+++ b/openquake/commonlib/source.py
@@ -413,15 +413,19 @@ class CompositeSourceModel(collections.Sequence):
                 logging.info('Creating %d realization(s) for model '
                              '%s, %s', len(rlzs), smodel.name, smodel.path)
                 idx = assoc._add_realizations(idx, smodel, rlzs)
-        # sanity check
+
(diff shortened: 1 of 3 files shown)


openquake/commonlib/source.py
```diff
@@ -396,6 +413,19 @@ def get_rlzs_assoc(self, get_weight=lambda tm: tm.num_ruptures):
...
+                assert len(assoc.realizations) == num_samples
+                for rlz in assoc.realizations:
+                    rlz.weight = 1. / num_samples
+            else:
+                tot_weight = sum(rlz.weight for rlz in assoc.realizations)
+                if tot_weight < 1:
+                    logging.warn('Some source models are not contributing, '
+                                 'weights are being rescaled')
+                for rlz in assoc.realizations:
+                    rlz.weight = rlz.weight / tot_weight
```
ptormene: We can be confident that tot_weight is never zero here, right?

PR #160, opened 2015-02-25
Manage properly the oversampling situation

micheles:
See https://bugs.launchpad.net/oq-engine/+bug/1425030.
The tests are green: https://ci.openquake.org/job/zdevel_oq-risklib/322


openquake/commonlib/source.py
```diff
@@ -394,6 +407,13 @@ def get_trt_id(self, col_id):
         raise KeyError('There is no TrtModel associated to the collection %d!'
                        % col_id)
 
+    def get_triples(self):
+        """
+        Yield triples (trt_id, idx, col_id) in order
```
ptormene: Please rename col_id to col_idx also here

openquake/commonlib/source.py
```diff
@@ -454,24 +474,20 @@ def get_rlzs_assoc(self, get_weight=lambda tm: tm.num_ruptures):
...
-            else:
-                tot_weight = sum(rlz.weight for rlz in assoc.realizations)
-                if tot_weight == 0:
-                    raise ValueError('All realizations have zero weight??')
-                elif tot_weight < 1:
-                    logging.warn('Some source models are not contributing, '
-                                 'weights are being rescaled')
-                for rlz in assoc.realizations:
-                    rlz.weight = rlz.weight / tot_weight
+        if num_samples:
```
ptormene: Please check if there's actually any difference between the current situation and the old one

openquake/commonlib/source.py
micheles: Done

openquake/commonlib/source.py
micheles: There is no real difference. I am reverting to the original.

openquake/commonlib/calculators/classical.py
```diff
@@ -22,6 +22,8 @@
 from functools import partial
 import numpy
 
+import numpy
```
ptormene: numpy was already imported

micheles: Ops, this was a merge error

openquake/commonlib/readinput.py
```diff
@@ -414,11 +412,9 @@ def get_source_models(oqparam, source_model_lt, sitecol=None):
                 trt_model.gsims = gsim_lt.values[trt_model.trt]
         else:
             gsim_lt = logictree.DummyGsimLogicTree()
-        weight = (rlz.weight if oqparam.calculation_mode == 'event_based'
-                  else rlz.weight / num_samples)
+        weight = rlz.weight / num_samples
```
ptormene: We might add a check in case num_samples is zero

ptormene:

openquake/commonlib/readinput.py
micheles: It cannot happen. num_samples is built by counting the source model realizations, so at minimum it is 1.

PR #161, opened 2015-02-25
Add a reduce_exposure.py script

micheles:
See https://bugs.launchpad.net/oq-risklib/+bug/1425561.
The tests are green: https://ci.openquake.org/job/zdevel_oq-risklib/299

PR #163, opened 2015-02-27
Fixed the management of leaf nodes

micheles:
This solves the problem with the current engine master by fixing an error introduced in https://github.com/gem/oq-risklib/pull/161

The tests are running here: https://ci.openquake.org/job/zdevel_oq-risklib/302


ptormene:
LGTM

PR #166, opened 2015-03-02
Plotting hazard curves

micheles:
and other improvements. See https://bugs.launchpad.net/oq-engine/+bug/1426821.
The tests are green: https://ci.openquake.org/job/zdevel_oq-risklib/313/


ptormene:
Apart from the legend overlapping with the plots, and from the not intuitive error messages in case nothing can be plotted, LGTM

micheles:
I have added a nicer error message for the case when no hazard curves are stored in the .pik file.

PR #167, opened 2015-03-02
Changed the expected event_loss_asset.csv

micheles:
Now zeros are not saved anymore: https://bugs.launchpad.net/oq-engine/+bug/1427226

Moreover, the ordering has been changed. Now it is first by `rupture_tag` and then by `asset_id`.


raoanirudh:
LGTM

PR #170, opened 2015-03-05
Added a debugging facility to reduce large source model

micheles:
The model is reduced by sampling randomly a subset of the sources. The code also work for the exposure.


ptormene:
LGTM

PR #184, opened 2015-03-20
Skipped the EventLossTestCase

micheles:
Temporary solution for the breakage in https://ci.openquake.org/job/master_oq-risklib/1137/

It seems to be one of those tests that sometimes pass and sometimes not.


daniviga:
LGTM

PR #187, opened 2015-03-23
Added an utility to pick up the realizations closest to the mean

micheles:
See https://bugs.launchpad.net/oq-engine/+bug/1429825. I am also adding a memory-efficient ArrayDict class that will be much used in the future.


raoanirudh:
LGTM

micheles:
Anirudh, I did expose the `min_value` parameter, as you suggested.

raoanirudh:
Re: pick_rlzs — it would be nice to have an option to set the minimum poe via the CLI rather than having it hardcoded. Apart from that, LGTM.

Re: ArrayDict — switching to a non-standard data structure would make the code slightly more inaccessible. Thus it would help to document somewhere in which cases one would prefer to use an ArrayDict over a nested dictionary. Apart from this comment, LGTM.

PR #189, opened 2015-03-24
Adds support for "Backarc" parameter

g-weatherill:
Implementation of the Abrahamson et al. (2015) GMPE requires the definition of a new site parameter ("backarc") (https://github.com/gem/oq-hazardlib/pull/308). This PR propagates the necessary changes into the risklib to allow the parameter to be read from i) the job.ini file, and ii) the site_model.xml

### DIFF
--- a/openquake/commonlib/tests/readinput_test.py
+++ b/openquake/commonlib/tests/readinput_test.py
@@ -61,6 +61,7 @@ export_dir = %s
                 'calculation_mode': 'classical',
                 'truncation_level': 0.0,
                 'random_seed': 0,
+                'reference_backarc': False,
                 'maximum_distance': 1.0,
                 'inputs': {'job_ini': job_config,
                            'site_model': site_model_input},
@@ -71,6 +72,8 @@ export_dir = %s
 
             with mock.patch('logging.warn') as warn:
                 params = vars(readinput.get_oqparam(job_config))
+                print params
+                print expected_params
                 self.assertEqual(expected_params, params)
                 self.assertEqual(['site_model', 'job_ini'],
                                  params['inputs'].keys())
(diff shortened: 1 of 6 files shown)


openquake/commonlib/tests/readinput_test.py
```diff
@@ -159,20 +163,20 @@ def test_get_site_model(self):
 <nrml xmlns:gml="http://www.opengis.net/gml"
       xmlns="http://openquake.org/xmlns/nrml/0.4">
     <siteModel>
-        <site lon="0.0" lat="0.0" vs30="1200.0" vs30Type="inferred" z1pt0="100.0" z2pt5="2.0" />
-        <site lon="0.0" lat="0.1" vs30="600.0" vs30Type="inferred" z1pt0="100.0" z2pt5="2.0" />
-        <site lon="0.0" lat="0.2" vs30="200.0" vs30Type="inferred" z1pt0="100.0" z2pt5="2.0" />
+        <site lon="0.0" lat="0.0" vs30="1200.0" vs30Type="inferred" z1pt0="100.0" z2pt5="2.0" backarc="False" />
+        <site lon="0.0" lat="0.1" vs30="600.0" vs30Type="inferred" z1pt0="100.0" z2pt5="2.0" backarc="False" />
+        <site lon="0.0" lat="0.2" vs30="200.0" vs30Type="inferred" z1pt0="100.0" z2pt5="2.0" backarc="False" />
     </siteModel>
```
micheles: I would set one of those to backarc="True". Then we can check that the flag is really read and that the test is not just using the default value.

openquake/commonlib/valid.py
```diff
@@ -670,21 +670,22 @@ def ab_values(value):
 vs30_type = ChoiceCI('measured', 'inferred')
 
 SiteParam = collections.namedtuple(
-    'SiteParam', 'z1pt0 z2pt5 measured vs30 lon lat'.split())
+    'SiteParam', 'z1pt0 z2pt5 measured vs30 backarc lon lat'.split())
```
micheles: Let's move backarc to the last position, to be consistent with the signature of the function site_param below.

micheles:
LGTM

PR #197, opened 2015-04-01
Extract the known oqparams from the class, not from the instance

micheles:
Fix the error in https://ci.openquake.org/job/master_oq-engine/1652/


daniviga:
LGTM

PR #206, opened 2015-04-09
Integrate the oq-lite calculators into the engine: phase 1

micheles:
See https://bugs.launchpad.net/oq-engine/+bug/1442013. The tests are running here: https://ci.openquake.org/job/zdevel_oq-risklib/454

PR #207, opened 2015-04-10
Restored filtering

micheles:
Restored the source filtering as it was (it was temporarily removed in https://github.com/gem/oq-risklib/pull/206). The tests are green: https://ci.openquake.org/job/zdevel_oq-risklib/457/


ptormene:
LGTM

PR #211, opened 2015-04-14
Fixed the errors in https://ci.openquake.org/job/master_oq-risklib/1193/

micheles:
The changes in the epsilon management changed slightly the expected numbers. It seems that there is a block size dependency, so for the moment I am hard-coding the number of blocks to 32, until I do not find the culprit. The tests are now passing: https://ci.openquake.org/job/zdevel_oq-risklib/467


ptormene:
LGTM

PR #222, opened 2015-04-25
Added a nontrivial test case for the event based calculator

micheles:
This is a case where the logic tree is non trivial, both for the sources and the GMPEs. It has 2 source models, each one with 2 tectonic region types and 4  gsim realizations. The test is running here: https://ci.openquake.org/job/zdevel_oq-risklib/508/


openquake/commonlib/tests/calculators/event_based_risk_test.py
```diff
@@ -33,3 +32,11 @@ def test_case_2(self):
         self.assertEqualFiles(
             'expected/rlz-000-structural-loss-curves.csv',
             out['rlz-000-structural-loss-curves'])
+
+    @attr('qa', 'risk', 'event_loss')
```
ptormene: Please align names (event_loss or event_based_risk)

ptormene:
Apart from the renaming, LGTM

PR #227, opened 2015-04-29
Extended the multiple gsim support to scenario_risk and scenario_damage

micheles:
See https://bugs.launchpad.net/oq-engine/+bug/1449833. The tests are green: https://ci.openquake.org/job/zdevel_oq-risklib/527


openquake/commonlib/calculators/scenario_risk.py
```diff
@@ -124,8 +125,21 @@ def pre_execute(self):
 
     def post_execute(self, result):
         """
-        Export the aggregate loss curves in CSV format.
+        Export the loss curves in CSV format.
```
raoanirudh: Export the mean and standard deviation of losses in CSV format.

raoanirudh:
The two calculators fail as expected when a GMPE logic tree with more than one TRT is used. It would be nice to provide a more informative error message for this.

Apart from that LGTM!

micheles:
Actually you have found in the bug in the procedure reporting the informative error message! So I am adding a test for the error situation too.

PR #228, opened 2015-04-29
Restored the ability to parallelize simple functions

micheles:
Now oq-lite is able to parallelize regular functions again. That means that you do no need to use any decorator. The tests are running here: https://ci.openquake.org/job/zdevel_oq-risklib/528


ptormene:
LGTM

PR #231, opened 2015-04-29
Restored a default for the parameter risk_imtls

micheles:
This should not be needed, but let's see if it solves the problem in Jenkins: https://ci.openquake.org/job/master_oq-risklib/1232/

zdevel is green: https://ci.openquake.org/job/zdevel_oq-risklib/529


daniviga:
LGTM and cross fingers

PR #246, opened 2015-05-19
Scenario damage bug

micheles:
See https://bugs.launchpad.net/oq-engine/+bug/1456431. The tests are running here: https://ci.openquake.org/job/zdevel_oq-risklib/616/


raoanirudh:
The exporter bug fix looks good to me. Apart from this fix, we will need to add a test case to check the numbers when there are multiple assets at a node.

micheles:
Yes, we will add the test in another pull request. Let's close this so the nightly builds of tomorrow will work correctly.

PR #249, opened 2015-05-20
Improved the `info` command to show information about the input/output weights

micheles:
This is useful to know how big is a computation *before* running it. User of OATS may check if the computation is outside the weight limits. Here is an example of usage:

```bash
$ oq-lite info job.ini -f
Reading and filtering the source model...
<CompositionInfo
b1, NSHMP2008US.xml, trt=[0]: 1 realization(s)>
<RlzsAssoc
0,ChiouYoungs2008: ['<0,b1,b3,w=1.0>']>
input_weight 14504
max_realizations 1
n_imts 1
n_levels 19.0
n_sites 1000
output_weight 19000.0
```

### DIFF
--- a/openquake/commonlib/commands/info.py
+++ b/openquake/commonlib/commands/info.py
@@ -46,19 +46,16 @@ def info(name, filtersources=False):
                     oqparam), []
         else:
             sitecol, assets_by_site = None, []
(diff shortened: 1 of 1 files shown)


daniviga:
LGTM

PR #251, opened 2015-05-21
Several improvements to the event_based_risk calculator

micheles:
Work to fix https://bugs.launchpad.net/oq-engine/+bug/1457463. It includes several new things:

  + cleanup method in the base calculator
  + counts_asset method
  + initial support for the counts matrix
  + hack for the specific_assets
  + exported avg_losses and agg_avg_losses
  + warning if `ground_motion_fields` is not honored
  + changed the names of some datasets in the datastore
  + improved the commands `run` and `show`
  + now we get the right byte count for HDF5 objects too
  + fixed the ordering of the sites when coming from the site model
  + fixed the mismatch hazard_investigation_time <-> risk_investigation_time
  + changed the risk_investigation_time in event_based_risk/case_3
  + now the assets are sent to the workers only once

The tests are green: https://ci.openquake.org/job/zdevel_oq-risklib/640/


openquake/commonlib/calculators/base.py
```diff
@@ -70,7 +71,8 @@ def __init__(self, oqparam, monitor=DummyMonitor(), calc_id=None,
         else:
             self.datastore = general.AccumDict()
             self.datastore.hdf5 = {}
-        self.oqparam = oqparam
+        if 'oqparam' not in self.datastore:  # new datastore
+            self.oqparam = oqparam
```
ptormene: Please check this

micheles: I checked. This is actually correct. I have added a comment explaining that this is the case when there is a precalculator and oqparam has been already stored in the datastore.

openquake/commonlib/calculators/event_based_risk.py
```diff
@@ -74,19 +80,37 @@ def event_based_risk(riskinputs, riskmodel, rlzs_assoc, monitor):
...
     return names
 
 
+def extract_avglosses(dstore, pattern):
+    """
+    Function extracting the average losses from the loss curves
+    for each realization.
+
+    :param dstore: a datastore object
+    :param pattern: '/loss_curves-rlzs/%s' or '/agg_loss_curve-rlzs/%s'
```
ptormene: Do we need a check that pattern is actually like this?

openquake/commonlib/calculators/event_based_risk.py
```diff
@@ -151,25 +174,28 @@ def post_execute(self, result):
         several interesting outputs.
         """
         oq = self.oqparam
+        self.rlzs_assoc = self.rlzs_assoc  # write anew
```
ptormene: I don't understand this line

openquake/commonlib/calculators/event_based_risk.py
```diff
@@ -204,7 +234,9 @@ def post_execute(self, result):
                         data_by_lt[loss_type].append(
                             (tag, asset_id, loss, ins_loss))
                 for loss_type, data in data_by_lt.iteritems():
-                    event_loss_asset[i][loss_type] = sorted(data)
+                    event_loss_asset[i][loss_type] = sorted(
+                        (t, a, l, i) for t, a, l, i in data
```
ptormene: Please add a comment

ptormene:
Besides the comments, LGTM

openquake/commonlib/calculators/event_based_risk.py
micheles: If not, one immediately gets a clear KeyError from a couple of lines below. So, it is okay as it is, IMO.

openquake/commonlib/calculators/event_based_risk.py
micheles: It takes the attribute `self.rlzs_assoc` from the cache (where it may have been updated) and write it again on the datastore.

openquake/commonlib/calculators/event_based_risk.py
micheles: Added

PR #253, opened 2015-05-21
Extend the XML representation for SimpleFaultSources to include the parameters needed for the directivity model

micheles:
See https://bugs.launchpad.net/oq-engine/+bug/1457489.
The tests are running here: https://ci.openquake.org/job/zdevel_oq-risklib/631


openquake/commonlib/nrml_examples/source_model/mixed.xml
```diff
@@ -123,6 +123,22 @@ xmlns:gml="http://www.opengis.net/gml"
...
+                weight="0.7"
+                >
+                    90.0
+                </slip>
+                <slip
+                weight="0.3"
+                >
+                    135.0
+                </slip>
+            </slipList>
```
mmpagani: I find the format adopted for the two list not fully consistent, but I'm sure you did it for specific reasons.

openquake/commonlib/nrml.py
```diff
@@ -159,6 +159,9 @@ class ValidNode(LiteralNode):
         nodalPlane=valid.nodal_plane,
         occurRates=valid.positivefloats,
         probs_occur=valid.pmf,
+        weight=valid.probability,
+        alongStrike=valid.probability,
+        downDip=valid.probability,
```
mmpagani: Can you check if there's a validation for slip? I don't see it here.

micheles: Look at the function `def slip_list(nodes)`, the validation is there (there is a slip_range from 0 to 360 degrees).

openquake/commonlib/nrml_examples/source_model/mixed.xml
micheles: I borrowed the format for the `hypo` nodes from the `hypocenter` nodes, but perhaps it was not a good idea. What would be a more consistent format? Perhaps

```
<hypo weight="0.3"/>
0.25 0.25
</hypo>
```

or
```
<hypo weight="0.3"/>
<alongStrike>0.25</alongStrike>
<downDip>0.25</downDip>
</hypo>
```
?

openquake/commonlib/tests/source_test.py
```diff
@@ -477,6 +479,65 @@ def test_raises_useful_error_2(self):
...
+            </rake>
+            <hypoList>
+                <hypo alongStrike="0.25" downDip="0.25" weight="0.3"/>
+                <hypo alongStrike="0.75" downDip="0.75" weight="0.7"/>
+            </hypoList>
+        </simpleFaultSource>
+    </sourceModel>
+</nrml>
+""")
+        # check that the error raised by hazardlib is wrapped correctly
```
mmpagani: This is more a question for my hazard colleagues. Why are we assuming that once we have a hypocenter distribution we necessarily need a slid distribution? This is a test we should have when we compute directivity but not necessarily here.

mmpagani:
LGTM pending on the comments/points provided above. Thank you all.

openquake/commonlib/tests/source_test.py
vup1120: Ciao, Marco, yes, I agree. It should be a better idea not to make the hypocentre distribution and slip distribution explicitly used for directivity calculation. Maybe tomorrow we can discuss how to put this test for directivity calculation.

micheles:
After discussion with you this morning, I agree that the check on the hypoList-splipList could be removed, but this is done in hazardlib and it is independend from this pull request. So I am going to merge.

PR #255, opened 2015-05-25
Added an ID to the VulnerabilityFunction class

micheles:
This is needed in order to be able to serialize as XML a VulnerabilityFunction object.
The tests are running here: https://ci.openquake.org/job/zdevel_oq-risklib/644

### DIFF
--- a/openquake/risklib/qa_tests/event_based_test.py
+++ b/openquake/risklib/qa_tests/event_based_test.py
@@ -47,7 +47,7 @@ class EventBasedTestCase(unittest.TestCase):
         # This is a regression test. Data has not been checked
         vf = (
             scientific.VulnerabilityFunction(
-                'PGA',
+                'SOME-TAXONOMY', 'PGA',
                 [0.001, 0.2, 0.3, 0.5, 0.7],
                 [0.01, 0.1, 0.2, 0.4, 0.8],
                 [0.01, 0.02, 0.02, 0.01, 0.03]))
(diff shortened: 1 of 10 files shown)


openquake/risklib/qa_tests/event_based_test.py
```diff
@@ -88,7 +88,7 @@ def test_mean_based_with_partial_correlation(self):
         # This is a regression test. Data has not been checked
         vf = (
             scientific.VulnerabilityFunction(
-                'PGA',
+                'SOME-TAXONOMY', 'PGA',
```
raoanirudh: workflows.ProbabilisticEventBased on line 106 uses `SOMETAXONOMY` rather than `SOME-TAXONOMY`.
Is that relevant?

openquake/risklib/qa_tests/event_based_test.py
```diff
@@ -127,7 +127,7 @@ def test_mean_based_with_perfect_correlation(self):
...
```
raoanirudh: Same comment as above

openquake/risklib/qa_tests/event_based_test.py
micheles: No, because the ID is ignored in such tests, however, let me fix this, to avoid any possible confusion.

openquake/risklib/qa_tests/event_based_test.py
```diff
@@ -212,7 +212,7 @@ def test_mean_based(self):
 
     def test_insured_loss_mean_based(self):
         vf = scientific.VulnerabilityFunction(
-            'PGA',
+            'VF', 'PGA',
```
raoanirudh: Same comment as above

raoanirudh:

PR #261, opened 2015-05-28
Making sure that everything works even when there are more hazard sites than risk sites

micheles:
See https://bugs.launchpad.net/oq-engine/+bug/1457062.
The tests are green: https://ci.openquake.org/job/zdevel_oq-risklib/663/


ptormene:
LGTM

PR #271, opened 2015-06-09
Improved the show command and added a purge command

micheles:
The demos are running here: https://ci.openquake.org/job/zdevel_oq-risklib/702/console

### DIFF
--- a/openquake/commonlib/commands/show.py
+++ b/openquake/commonlib/commands/show.py
@@ -36,7 +36,7 @@ def show(calc_id, key=None, rlzs=None):
     if not calc_id:
         if not os.path.exists(datastore.DATADIR):
             return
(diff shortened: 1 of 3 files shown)


openquake/commonlib/commands/show.py
```diff
@@ -30,19 +33,31 @@ def show(calc_id, key=None, rlzs=None):
     :param key: key of the datastore
     :param rlzs: flag; if given, print out the realizations in order
     """
+    if not calc_id:
```
ptormene: Please add to the docstring the fact that calc_id==0 means all (as in purge)

ptormene:
Besides the minor comment, LGTM

openquake/commonlib/commands/show.py
micheles: Done

PR #272, opened 2015-06-09
Store the epsilons matrix in HDF5

micheles:
See https://bugs.launchpad.net/oq-engine/+bug/1448932.
The tests are running here: https://ci.openquake.org/job/zdevel_oq-risklib/704

You can run a calculation and open the output.hdf5 file, or export it to CSV with the command

```bash
$ oq-lite export -1 /epsilon_matrix
```


raoanirudh:
It would be nice if we could include a column with the asset refs. Other than that, LGTM.

micheles:
Actually that column is not stored on purpose.

PR #276, opened 2015-06-13
Restored the oq-lite tiling calculator

micheles:
I have restored a test that was temporarily skipped and changed the calculator so that the workers do not need access to the file system, so that https://bugs.launchpad.net/oq-engine/+bug/1400041 is closed, at least in principle (then we will need to optimize for large calculations for sure).
Tests are green: https://ci.openquake.org/job/zdevel_oq-risklib/721.

### DIFF
--- a/openquake/commonlib/calculators/classical.py
+++ b/openquake/commonlib/calculators/classical.py
@@ -189,7 +189,7 @@ def is_effective_trt_model(result_dict, trt_model):
 
 
 @parallel.litetask
-def classical_tiling(calculator, sites, indices, tileno, monitor):
+def classical_tiling(calculator, sites, position, tileno, monitor):
     """
     :param calculator:
         a ClassicalCalculator instance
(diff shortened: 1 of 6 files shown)


openquake/commonlib/calculators/classical.py
```diff
@@ -190,14 +189,16 @@ def is_effective_trt_model(result_dict, trt_model):
...
         a ClassicalCalculator instance
     :param sitecol:
-        a SiteCollection instance
-    :param tileno:
-        the number of the current tile
+        the site collection of the current tile
+    :param indices:
+        the indices of the sites in the current tile
+    :param position:
+        position of the current tile in the full site collection
```
ptormene: Plese fix the docstring

ptormene:
Apart from the docstring, LGTM

openquake/commonlib/calculators/classical.py
micheles: Done

PR #285, opened 2015-06-17
Store the realizations in the HDF5 file

micheles:
As requested by Marco, see https://bugs.launchpad.net/oq-engine/+bug/1465624. The tests are running here: https://ci.openquake.org/job/zdevel_oq-risklib/739

### DIFF
--- a/openquake/commonlib/calculators/base.py
+++ b/openquake/commonlib/calculators/base.py
@@ -42,6 +42,8 @@ calculators = general.CallableDict(operator.attrgetter('calculation_mode'))
 class AssetSiteAssociationError(Exception):
     """Raised when there are no hazard sites close enough to any asset"""
 
+rlz_dt = numpy.dtype([('uid', (str, 200)), ('weight', float)])
+
 
 class BaseCalculator(object):
     """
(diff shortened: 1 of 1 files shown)


openquake/commonlib/calculators/base.py
```diff
@@ -42,6 +42,8 @@
 class AssetSiteAssociationError(Exception):
     """Raised when there are no hazard sites close enough to any asset"""
 
+rlz_dt = numpy.dtype([('uid', (str, 200)), ('weight', float)])
```
mmpagani: I'm not a python coder by I find the position of this statement a bit confusing

micheles: I am defining a new datatype, with static parameters, this is done usually at top level, no?

mmpagani:
This was a very easy code review :D. I think we will need to explain somewhere that rlz-0 corresponds to the first path in the 'realisations' attribute.
LGTM

For the future:
I find the current representation of the path a bit unclear since the branches IDs are not unique and can be repeated in different branching levels. I would find more intuitive an ID for each branch which is a combination of the id of the branching level (an integer) and the branch ID (another integer) e.g. 01:01

PR #287, opened 2015-06-19
Added a facility to create dataset directly from the datastore

micheles:
The Hdf5Dataset clas is wrapping the resize logic of HDF5 datasets, which is a bit annoying. In particular it manages the position bookkeeping for extendable datasets. The tests are running here:
https://ci.openquake.org/job/zdevel_oq-risklib/741


ptormene:
LGTM

PR #290, opened 2015-06-22
Implemented vulnerability functions with PMF

micheles:
See https://bugs.launchpad.net/oq-engine/+bug/1459087. The tests are running here: https://ci.openquake.org/job/zdevel_oq-risklib/752/


openquake/commonlib/nrml.py
```diff
@@ -313,12 +315,13 @@ def read(source):
     """
     nrml = parse(source).getroot()
     assert striptag(nrml.tag) == 'nrml', nrml.tag
+    xmlns = nrml.tag.split('}')[0][1:]
```
micheles: Add a comment here, the line is a bit cryptic

micheles:
LGTM, but we must check the numbers on the engine side

PR #291, opened 2015-06-23
Support for uniform_hazard_spectra = true and hazard_maps = false

micheles:
In such case the hazard maps must be computed but not saved in the datastore. See the output of https://ci.openquake.org/job/zdevel_oq-risklib/759/console


ptormene:
LGTM

PR #292, opened 2015-06-23
Recognize negative calculation IDs

micheles:
Now you can do

```bash
oq-lite run job_hazard.ini
oq-lite run job_risk.ini --hc -1
```

See https://ci.openquake.org/job/zdevel_oq-risklib/760/console


raoanirudh:
LGTM

PR #301, opened 2015-06-27
Fixed writers.write_csv

micheles:
The write_csv utility was saving composite numpy arrays into .csv files without preserving the order of the fields. The fix is to change `dtype.fields -> dtype.names. The tests are running here:
https://ci.openquake.org/job/zdevel_oq-risklib/782/


ptormene:
LGTM

PR #304, opened 2015-06-29
Replace the oq-lite script with an entry point

micheles:
This is the cleanest and most modern thing to do, when installing a Python library. It also avoids the forking problem that was affecting Windows. See https://bugs.launchpad.net/oq-engine/+bug/1469616, the tests are green: https://ci.openquake.org/job/zdevel_oq-risklib/790/


daniviga:
LGTM

PR #305, opened 2015-06-29
setuptools is now required even after the build

micheles:
This fixes the error in https://ci.openquake.org/job/master_oq-risklib/1391/
The reason is that oq-lite is now an entry point.
The demos are running now: https://ci.openquake.org/job/zdevel_oq-risklib/792

NB: on mac/linux it is convenient to define oq-lite as a symlink to `oq-risklib/openquake/commonlib/commands/__main__.py`

### DIFF
--- a/bin/oq-lite
+++ b/bin/oq-lite
@@ -1,0 +1,1 @@
+../openquake/commonlib/commands/__main__.py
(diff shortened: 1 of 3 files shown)


daniviga:
LGTM

PR #309, opened 2015-07-02
Removed `investigation_time` from risk configuration files

micheles:
See https://bugs.launchpad.net/oq-engine/+bug/1470134
The tests are running here: https://ci.openquake.org/job/zdevel_oq-risklib/812


openquake/commonlib/oqvalidation.py
```diff
@@ -75,9 +75,8 @@ class OqParam(valid.ParamSet):
     intensity_measure_types_and_levels = valid.Param(
         valid.intensity_measure_types_and_levels, None)
     # hazard_imtls = valid.Param(valid.intensity_measure_types_and_levels, {})
-    hazard_investigation_time = valid.Param(valid.positivefloat, None)
     interest_rate = valid.Param(valid.positivefloat)
-    investigation_time = valid.Param(valid.positivefloat, None)
+    investigation_time = valid.Param(valid.positivefloat, 50.)
```
raoanirudh: It might be better to have no default value for `investigation_time` and instead throw an error if it is missing in the job file.

micheles: The issue is with scenario calculations, where the user should not set an investigation time. Then without a default an error would be raised and the user would be forced to add an otherwise unused value every time.

raoanirudh: The missing `investigation_time` error should be raised only if the calculation mode is not scenario. Using a default value when the user forgets to include this parameter in the job file could be misleading.

micheles: I see your point. Let me try to see what happens if I remove the default.

raoanirudh:
LGTM from the risk POV.

PR #312, opened 2015-07-03
Fixed the expected hazard curves

micheles:
By mistake I did commit the hazard curves computed with investigation_time=100 instead of the right ones with investigation_time=10. This is the reason why https://ci.openquake.org/job/master_oq-risklib/1400/ is broken.


ptormene:
LGTM

PR #314, opened 2015-07-06
MANIFEST.in: missing data from qa_tests_data

daniviga:
Otherwise data needed to run the `qa_tests` is not included in package

Test: https://ci.openquake.org/job/zdevel_oq-risklib/831/


micheles:
If the tests pass, LGTM.

PR #321, opened 2015-07-08
Added reports in rst format for all tests and demos

micheles:
This is the second part of https://bugs.launchpad.net/oq-engine/+bug/1471648. No tests are needed since there are no code changes.


daniviga:
LGTM

PR #329, opened 2015-07-18
Using 100 SES instead of 200 to save memory

micheles:
Master is currently broken because of an OutOfMemory error in the ProbabilisticEventBasedDemo:
https://ci.openquake.org/job/master_oq-engine/1879/console

I am reducing the memory consumption so that the demo run again (see
https://ci.openquake.org/job/zdevel_oq-engine/1355/).  Further investigation will be needed when I will come back from my vacation. This is not a big issue since all tests are green.


micheles:
LGTM

PR #330, opened 2015-07-27
Using the ebr calculator in the ProbabilisticEventBased demo

micheles:
This saves a lot of memory, so I could restore the number of SES to 200.


daniviga:
LGTM

daniviga:
LGTM

PR #331, opened 2015-07-27
Support for both NRML 0.4 and NRML 0.5

micheles:
For the moment we want to support only the new vulnerability functions. In the future we will drop the support for the 0.4 format and force all users to upgrade. The conversion tool is already included and a lot of XML files have been already migrated. The tests are running here: https://ci.openquake.org/job/zdevel_oq-risklib/892/


raoanirudh:
LGTM

PR #350, opened 2015-08-25
Increased the export precision

micheles:
Internally lons and lats have 8 digits of precision, i.e. we can distinguish 179.00000 and 179.00001 (3 digits before the decimal point + 5 digits after. This level of precision should be kept when exporting.
See https://bugs.launchpad.net/oq-engine/+bug/1488418 and the tests https://ci.openquake.org/job/zdevel_oq-risklib/972/


raoanirudh:
LGTM

PR #351, opened 2015-08-26
Mean avg losses

micheles:
Added an utility view to display the mean average losses in a way consistent with the engine.
Also, fixed a bug in the GMF exporter. The tests are running here: https://ci.openquake.org/job/zdevel_oq-risklib/978


ptormene:
LGTM

PR #355, opened 2015-08-31
Restricted the valid source IDs

micheles:
See https://bugs.launchpad.net/oq-engine/+bug/1490490. The tests are green: https://ci.openquake.org/job/zdevel_oq-risklib/986

### DIFF
--- a/openquake/commonlib/valid.py
+++ b/openquake/commonlib/valid.py
@@ -160,6 +160,16 @@ name = Regex(r'^[a-zA-Z_]\w*$')
 name_with_dashes = Regex(r'^[a-zA-Z_][\w\-]*$')
 
 
+def source_id(value):
+    """
+    Check the source id; the only accepted chars are a-zA-Z0-9_-
+    """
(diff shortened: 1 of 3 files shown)


openquake/commonlib/valid.py
```diff
@@ -159,6 +159,21 @@ def __call__(self, value):
...
+    """
+    Check the source id; the only accepted chars are a-zA-Z0-9_-
+    """
+    if len(value) > MAX_ID_LENGTH:
+        raise ValueError('The ID %r is longer than %d character' %
+                         (value, MAX_ID_LENGTH))
+    if re.match(r'^[\w_\-]+$', value):
+        return value
+    raise ValueError(
+        'The ID %r contains invalid characters' % value)
```
g-weatherill: It would be better for the error message to indicate explicitly which characters are valid (i.e. a-z A-Z 0-9 _ and -)

g-weatherill:
LGTM

PR #371, opened 2015-09-14
Removed an unwanted warning

micheles:
NRML 0.4 is not deprecated yet, see https://bugs.launchpad.net/oq-engine/+bug/1494742


ptormene:
LGTM

PR #380, opened 2015-09-16
Optimization on the average losses computation

### LINKED ISSUE #378: The sum of average losses is slow in the event based calculator
An approach which has proven spectacularly effective in the past is to replace the AccumDict with a numpy array.

micheles:
By replacing dictionaries with arrays we can improve a lot the aggregation time on the controller node.
Fixes #378; the tests are green: https://ci.openquake.org/job/zdevel_oq-risklib/1043/

PS: I am also trapping the case in which a symlink to the datastore cannot be made. This is not a problem since the symlinks are not used by the engine in any way, they are just an undocumented convenience for Anirudh.


ptormene:
LGTM

PR #398, opened 2015-09-25
Set concurrent_tasks=16 in demos/ProbabilisticEventBased

micheles:
To reduce the data transfer on Jenkins, currently 9.19 GB


daniviga:
LGTM

PR #401, opened 2015-09-28
Better printing of arrays

micheles:
Now ``oq-lite show <calc_id> <array_name>`` also prints an header with the column types. The tests are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1069


ptormene:
If the test is green, LGTM

PR #402, opened 2015-09-28
Load GMFs from a NRML file

micheles:
This is a feature that was missing from oq-lite. The tests are green: https://ci.openquake.org/job/zdevel_oq-risklib/1074/


openquake/calculators/base.py
```diff
@@ -505,20 +512,22 @@ def read_gmfs_from_csv(calc):
     :param calc: a ScenarioDamage or ScenarioRisk calculator
```
raoanirudh: It would be better to rename this function from `read_gmfs_from_csv` to `read_gmfs_from_file`, now that the gmfs can also be read from NRML files.

openquake/commonlib/readinput.py
```diff
@@ -912,14 +932,72 @@ def get_gmfs(oqparam, sitecol):
...
+    :returns:
+        a triple (sitecol, rupture_tags, gmf array)
+    """
+    imts = list(oqparam.imtls)
+    imt_dt = numpy.dtype([(imt, float) for imt in imts])
+    gmfset = nrml.read(fname).gmfCollection.gmfSet
+    tags, oqparam.sites = _extract_tags_sites(gmfset)
+    oqparam.number_of_ground_motion_fields = e = len(tags)
+    sitecol = get_site_collection(oqparam)
+    n = len(oqparam.sites)
```
raoanirudh: The variables `e`, `n`, `m`, `i`, `j` in this function could potentially be renamed to have more reader friendly names.

openquake/qa_tests_data/scenario_risk/case_2/gmfs.csv
```diff
@@ -0,0 +1,1000 @@
+scenario-000,,0.17111 0.305128 0.604032
```
raoanirudh: If the main purpose of this case is to test the reading of gmfs from a csv file, maybe we could reduce the number of gmfs to 5–10.

raoanirudh:
Reading gmfs from an NRML file works as expected. LGTM.

openquake/qa_tests_data/scenario_risk/case_2/gmfs.csv
micheles: That's a good idea but right now I do not want to do that, otherwise I would break the equivalent test on the engine.

openquake/calculators/base.py
micheles: Done

PR #404, opened 2015-09-30
Fixed an excessive validation

micheles:
This was breaking the engine: https://ci.openquake.org/job/master_oq-engine/2009/console


micheles:
LGTM

PR #406, opened 2015-10-01
Saved the scenario_risk outputs on the datastore

### LINKED ISSUE #400: Port the scenario_risk tests case_1, case_2 to oq-lite
Those are the tests where the GMFs are read from a .csv file. Also, save the outputs of the scenario_risk calculator on the datastore and add loss map export functionality.

micheles:
Closes #400. The tests are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1111


raoanirudh:
The directory specified with the export-output flag is not considered when exporting the avgloss or aggloss outputs. Apart from that, LGTM.

micheles:
The export is now fixed and I have also added the XML export for loss maps. Now the calculator is a full replacement for the engine one.

raoanirudh:
The aggloss numbers exported from the database and the datastore seem to be different.

micheles:
I confirm that there is a difference with the engine calculator in the case of the Italy model. However the difference is not related to this pull request, it was there already with the previous oq-lite calculator. So I propose to merge this PR (which is blocking other more important work) and to investigate the issue more deeply in a second moment. We cannot remove the old scenario_risk calculator until the problem is fully understood. It is not related to the epsilon sampling and cannot be seen in any of our existing tests.

raoanirudh:
Agreed, please go ahead with the merge.

PR #419, opened 2015-10-08
Simplified EventBasedRiskTestCase.test_case_3

micheles:
It is important to have a test which is fast and simple to debug, so I have simplified the test case_3 to work on a single asset. The test also exposes a corner case where the package h5py version 2.0.1
fails. We need a backport of h5py version 2.2.1 to have this run on Ubuntu 12.04. Fortunately we tested that we can keep the underlying HDF5 libraries as they are.


micheles:
LGTM on Matteo's part

nastasi-oq:
LGTM on Michele's part

Issue #421, opened 2015-10-09
Implement the logic for the consequence ratio


micheles:
See Anirudh's comment below.

raoanirudh:
Background and full specifications for the extensions to the Scenario Damage calculator to handle multiple loss types and calculate consequences. Specs also hosted at: https://gist.github.com/raoanirudh/a7fd17682fbc3db5538a 

## Introduction

The Scenario Damage Calculator computes damage distribution statistics for all assets in a given exposure model for a single specified earthquake rupture. Damage distribution statistics include the mean and standard deviation of damage fractions for different damage states. This calculator requires the definition of a finite rupture model, an exposure model and a fragility model; the main results are the damage distribution statistics per asset, aggregated damage distribution statistics per taxonomy, aggregated damage distribution statistics for the region, and collapse maps.

Damage distribution statistics can be computed for four different loss types using this calculator: structural damage, nonstructural damage, contents damage, and business interruption.

The rupture characteristics—i.e. the magnitude, hypocenter and fault geometry—are modelled as deterministic in the scenario calculators. Multiple realizations of different possible ground motion fields (GMFs) due to the single rupture are generated, taking into consideration both the inter-event variability of ground motions, and the intra-event residuals obtained from a spatial correlation model for ground motion residuals. The use of logic-trees allows for the consideration of uncertainty in the choice of a GMPE model for the given tectonic region.

As an alternative to computing the GMFs with OpenQuake, users can also provide their own sets of GMFs as input to the Scenario Damage Calculator.

For each GMF realization, damage fractions (the fraction of buildings in each damage state) are estimated for every asset in the exposure model using the provided fragility model, and finally the damage distribution statistics (i.e., the mean damage fractions and standard deviation of damage fractions for all damage states) across all realizations are calculated. The calculator also provides aggregated damage distribution statistics for the portfolio, such as mean damage fractions and standard deviation of damage fractions for each taxonomy in the exposure model, and the mean damage fractions and standard deviation of damage fractions for the total portfolio.

If a consequence model is also provided in addition to the fragility model, the damage state probabilities (i.e. the damage fractions) for each asset are multiplied by the consequence ratios for the corresponding damage states and summed to give the expected consequence ratio for each GM realization. The consequence ratio is multiplied by the respective value of the asset to give the expected consequence. Finally, the mean and standard deviation of the consequence are computed. Similar to the damage distribution, the aggregated consequences per taxonomy and for the total portfolio are also computed.

## Inputs

[...]

PR #431, opened 2015-10-14
Added a test for the consequences

micheles:
This is the test case_1h provided by Anirudh. Please check the numbers. This is another step needed for https://github.com/gem/oq-risklib/issues/421. The CSV export may be refined later on, for now I am exporting directly the HDF5 datasets.


openquake/qa_tests_data/scenario_damage/case_1h/fragility_model.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<nrml xmlns:gml="http://www.opengis.net/gml"
```
raoanirudh: Should we remove the gml line here too?

micheles: Yes, good catch.

raoanirudh:
The results for both the damage distribution and consequences for this single asset case match the expected values. LGTM!

PR #432, opened 2015-10-14
Fixed a pickle bug and added a multi-asset multi-gsim multi-loss-type test

micheles:
This makes it possible to run the test in https://github.com/gem/oq-risk-tests/pull/16


raoanirudh:
The test runs and all 27 of the expected output files are exported. LGTM.

micheles:
Actually the CSV export is correct even if scarcely readable. In this case there are two realizations (there are two GSIMs) and if notice carefully each column contains two floats, one for the first realization and one for the second realization. I am changing the exporter to to produce two different files in such situation.

raoanirudh:
LGTM

PR #434, opened 2015-10-15
Added a mechanism to support multiple NRML versions

micheles:
The approach used here costs less effort than deprecating NRML 0.4. I am also adding some .rst reports for the new tests that were forgotten.


ptormene:
If tests are green, LGTM

PR #437, opened 2015-10-15
Removed the symlink feature

micheles:
Anirudh was the only user of this undocumented feature and he requested me to remove it, since he does not use it anymore.


raoanirudh:
LGTM

Issue #440, opened 2015-10-16
Add a test on the average losses


micheles:
A bug entered such that the nonstructural losses in the ScenarioRisk demo where zero. While the bug is fixed now, we must cover better the case with multiple loss types.

PR #441, opened 2015-10-16
Fixed zeros in ScenarioRisk demos

micheles:
As reported by Catalina and Anirudh. We need to add a test: https://github.com/gem/oq-risklib/issues/440


raoanirudh:
LGTM

PR #444, opened 2015-10-19
Simplified a few job.ini files for the classical_damage tests

micheles:
This should simplify my life when porting those tests to NRML 0.5.


ptormene:
LGTM

PR #447, opened 2015-10-19
"Fixed" Travis

micheles:
This is not really fixed, there is no reason why it works now, in the same sense there was no understandable reason why it did not work before. Still, for the moment we can get green builds.


micheles:
LGTM for the moment

PR #450, opened 2015-10-20
Set the correct EventBasedRiskCalculator.pre_calculator

micheles:
As it was before, we were not computing the hazard curves, even if hazard_curves_from_gmfs was set.


daniviga:
LGTM

PR #453, opened 2015-10-20
Renamed ProbabilisticEventBased/job_risk.ini

micheles:
In this way the demo can be submitted to the Web UI.


daniviga:
Please fix also https://github.com/gem/oq-engine/blob/master/packager.sh#L528

daniviga:
LGTM

PR #455, opened 2015-10-20
Removed assetcol.csv

micheles:
The exporter for the asset collection has been removed in https://github.com/gem/oq-risklib/pull/452.
This should fix the error in https://ci.openquake.org/job/master_oq-risklib/1636/


daniviga:
LGTM is green

PR #456, opened 2015-10-21
Simplified a bit the logic for no damage limit

micheles:
We don't need to pass around the add_zero flag.


daniviga:
I see the comment `# TODO: this is complicated and perhaps wrong: check with Anirudh` but if it's blocking and tests are green than LGTM

micheles:
The comment should not worry you. We are consistent with the existing specs even if I am not sure that the existing specs are correct ;-)

PR #459, opened 2015-10-22
Switched to setup.py test

micheles:
The advantage is that we can customize how the tests are run. Moreover a lot of warnings that were hidden by nose are now visible.
I have also changed the h5py wheel to version 2.2.1 and now the segmentation fault has disappeared!


daniviga:
LGTM

PR #461, opened 2015-10-22
Renamed avg_losses -> loss_map

micheles:
As per Vitor's request. The tests and demos are green: https://ci.openquake.org/job/zdevel_oq-risklib/1141/


micheles:
This was a bad idea since the avg_losses are not loss maps.

Issue #451, opened 2015-10-20
The event based risk demo requires 8 GB to run


micheles:
In general the event based risk calculator is consuming too much memory. We should optimize the memory consumption in some way.

PR #466, opened 2015-10-23
Reduced the memory consumption in the event based risk calculator

micheles:
Aggregating by array instead that by list of arrays. In our demo the improvement is from ~900 KB per task to ~300 KB per task (pickled size of the task result). Moreover, I am reducing the data transfer back by removing big attributes from the monitor. The difference is from ~73 MB to ~300 KB per task.
Closes https://github.com/gem/oq-risklib/issues/451

### DIFF
--- a/openquake/calculators/event_based_risk.py
+++ b/openquake/calculators/event_based_risk.py
@@ -145,6 +145,10 @@ def event_based_risk(riskinputs, riskmodel, rlzs_assoc, monitor):
                 result[idx] = [sum(lst, AccumDict())]
         else:
             result[idx] = []
+
+    # NB: don't return back large amounts of data
+    delattr(monitor, 'assets_by_site')
+    delattr(monitor, 'eps_dict')
     return result
 
 
--- a/openquake/calculators/event_based_risk.py
+++ b/openquake/calculators/event_based_risk.py
@@ -56,7 +56,8 @@ def cube(O, L, R, factory):
 
 
 @parallel.litetask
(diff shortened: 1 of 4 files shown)


openquake/calculators/event_based_risk.py
```diff
@@ -71,10 +78,13 @@ def event_based_risk(riskinputs, riskmodel, rlzs_assoc, monitor):
         a single array of dtype elt_dt, or an empty list
     """
     lti = riskmodel.lti  # loss type -> index
-    specific_assets = monitor.oqparam.specific_assets
-    result = cube(
-        monitor.num_outputs, len(lti), len(rlzs_assoc.realizations), list)
-    for out_by_rlz in riskmodel.gen_outputs(riskinputs, rlzs_assoc, monitor):
+    L, R = len(lti), len(rlzs_assoc.realizations)
```
nastasi-oq: L, R, l, r, some more letter in var names, please.

micheles: There is a long tradition in scientific computing about using one-letter variable for integers. Also, the names are used in all other calculators, so I want to keep them. `r` is the loop index in `range(R)`, `l` is the loop index in `range(L)`.

nastasi-oq: Few understandable in any case.
Maybe a document with this kind of practices could help developers in the future.

nastasi-oq:
LGTM

PR #467, opened 2015-10-23
Adds "incrementalMFDAbsolute" option to source model logic tree

g-weatherill:
Waaaay back in time, the "modify_set_mfd" option was added to the EvenlyDisctretizedMFD class in the oq-hazardlib (https://github.com/gem/oq-hazardlib/pull/262). This allows the user to define explicit incremental MFD values to a source (or sources) within a logic tree. This PR propagates this functionality to the oq-risklib, adding a new epistemic uncertainty type to the source model logic tree class: "incrementalMFDAbsolute".

The uncertainty model used to describe this takes the form of a string: 
`min_mag,bin_width,rate_1 rate_2 ... rate_2`
where the minimum magnitude, bin width and rates list are comma sparated, but the rates themselves are whitespace delimited.

### DIFF
--- a/openquake/commonlib/logictree.py
+++ b/openquake/commonlib/logictree.py
@@ -380,10 +380,6 @@ class BranchSet(object):
             # source didn't pass the filter
             return
 
-        #if not isinstance(source.mfd, openquake.hazardlib.mfd.TruncatedGRMFD):
-            # source's mfd is not gutenberg-richter
-        #    return
-
         self._apply_uncertainty_to_mfd(source.mfd, value)
 
     def _apply_uncertainty_to_mfd(self, mfd, value):
@@ -402,6 +398,7 @@ class BranchSet(object):
 
         elif self.uncertainty_type == 'maxMagGRAbsolute':
             mfd.modify('set_max_mag', dict(value=value))
+
         elif self.uncertainty_type == 'incrementalMFDAbsolute':
             min_mag, bin_width, occur_rates = value
             mfd.modify('set_mfd', dict(min_mag=min_mag, bin_width=bin_width,
--- a/openquake/commonlib/logictree.py
+++ b/openquake/commonlib/logictree.py
@@ -801,8 +801,8 @@ class SourceModelLogicTree(BaseLogicTree):
             if len(mbr) == 3:
                 min_mag, bin_width, rates = mbr
                 try:
-                    rates = positivefloats(rates)
-                except:
+                    rates = valid.positivefloats(rates)
+                except ValueError:
                     rates = []
                 if _float_re.match(min_mag) and _float_re.match(bin_width) and\
                         len(rates):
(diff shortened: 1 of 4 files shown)


openquake/commonlib/logictree.py
```diff
@@ -792,6 +796,21 @@ def validate_uncertainty_value(self, node, branchset, value):
                 node, self.filename,
                 'expected a pair of floats separated by space'
             )
+        elif branchset.uncertainty_type == 'incrementalMFDAbsolute':
+            mbr = value.split(',')
+            if len(mbr) == 3:
+                min_mag, bin_width, rates = mbr
+                try:
+                    rates = positivefloat(rates)
```
micheles: Should this be `valid.positivefloat`? Also, I would change the `except` clause to `except ValueError`.
Also, please run flake8 on this file, I get some warnings.

micheles:
Should we add an end-to-end test, or change a demo so that this feature is used?

g-weatherill:
This feature is needed to implement the 2014 US National Seismic Hazard Maps, so I can put together a demo using this option with one of their files once that has been completed.

openquake/commonlib/logictree.py
micheles: positivefloats is still not imported, so this cannot work. You are getting a `NameError: name 'positivefloats' is not defined` but since you are using a catch all `except` you are not recognizing that.

micheles:
LGTM from the technological point of view.

openquake/commonlib/tests/logictree_test.py
```diff
@@ -476,6 +476,38 @@ def test_b_gr_relative_wrong_format(self):
         self.assertEqual(exc.message, 'expected single float value',
                          "wrong exception message: %s" % exc.message)
```
micheles: I forgot one thing Graeme: `exc.message` is deprecated. User `str(exc)` instead,

g-weatherill: When I change exc.message to str(exc) the test fails with this error:

```======================================================================
FAIL: test_b_gr_relative_wrong_format (openquake.commonlib.tests.logictree_test.SourceModelLogicTreeBrokenInputTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/GW/Documents/GEM/dev_risklib/oq-risklib/openquake/commonlib/tests/logictree_test.py", line 477, in test_b_gr_relative_wrong_format
    "wrong exception message: %s" % str(exc))
AssertionError: wrong exception message: filename '/var/folders/wc/gthl0y7j14dflw20_77bv7d00000gn/T/tmpwIEHcy.lt', line 15: expected single float value
----------------------------------------------------------------------```

micheles: Please disregard my comment, Graeme. I see now that we have a custom class LogicTreeError with a `.message` attribute, so it was correct as it was. This is not the deprecated attribute `.message` inherited from `Exception`.

g-weatherill: So I can merge?

micheles: For me yes, but I guess you also need a review from somebody in the scientific team.

Issue #468, opened 2015-10-24
Reduce the data transfer of the epsilons


micheles:
Two ideas: replace the epsilon dictionary with the epsilon matrix and reduce the epsilon precision to 32 bits.

PR #469, opened 2015-10-24
Refactored the epsilon management

micheles:
Closes https://github.com/gem/oq-risklib/issues/468. The demos are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1149/console

### DIFF
--- a/openquake/risklib/riskinput.py
+++ b/openquake/risklib/riskinput.py
@@ -287,7 +287,7 @@ class RiskInput(object):
             self.weight += len(assets)
         self.taxonomies = sorted(taxonomies)
         self.tags = None  # for API compatibility with RiskInputFromRuptures
-        self.eps = eps
+        self.eps = {} if eps is None else eps
 
     @property
     def imt_taxonomies(self):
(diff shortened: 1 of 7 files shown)


openquake/risklib/riskinput.py
```diff
@@ -290,14 +287,14 @@ def __init__(self, imt_taxonomies, hazard_by_site, assets_by_site,
             self.weight += len(assets)
         self.taxonomies = sorted(taxonomies)
         self.tags = None  # for API compatibility with RiskInputFromRuptures
-        self.eps_dict = eps_dict or {}
+        self.eps = {} if eps is None else eps
```
nastasi-oq: type inconsistency: if eps is None seems to me more correct pass an empty numpy array.

micheles: Indeed there was a confusion here. The class `RiskInput` wants an epsilon dictionary and not a matrix; it is the class `RiskInputFromRuptures` that wants an epsilon matrix. I have updated the docstrings and the name of the parameter accordingly.

nastasi-oq:
Tests green: https://ci.openquake.org/job/zdevel_oq-risklib/1149
LGTM

Issue #457, opened 2015-10-21
Make it possible to pass the random_seed to VulnerabilityFunctionWithPMF


micheles:
Now it is hard-code to 42 in riskmodels.py. Notice that a very strong seed dependency is expected.

PR #470, opened 2015-10-26
Set the seed for VulnerabilityFunctionWithPMF

micheles:
Closes https://github.com/gem/oq-risklib/issues/457. Now by changing the parameter `random_seed` in the job.ini file the seed dependency can be studied, whereas before the seed was hard-coded to 42.


raoanirudh:
LGTM

PR #472, opened 2015-10-26
Add a check on the total losses for scenario_risk

micheles:
And used it to check the results of the ScenarioRisk demo, i.e. the sum of the average losses, see https://github.com/gem/oq-risklib/issues/440. The demo is green: https://ci.openquake.org/job/zdevel_oq-risklib/1147

NB: while adding the test a bug in the case with two GSIMs was discovered and fixed.


openquake/commonlib/commands/show.py
```diff
@@ -21,6 +21,9 @@
 import os
 import shutil
 import logging
+
+import numpy
```
nastasi-oq: Why this import is required ?

micheles: It is not needed, it is a leftover. I am removing it.

nastasi-oq:
After a new green run of tests LGTM.

PR #474, opened 2015-10-27
Now the calculators scenario_risk and scenario_damage work with a single .ini file

micheles:
Closes https://github.com/gem/oq-engine/issues/1866. I have also changed the `gmfs.txt` files (which are used internally for tests) by adding an header contained the sites, as sorted pairs of (lon, lat).
I have changed the demos to use a single .ini file. The demos are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1163


raoanirudh:
LGTM

Issue #442, opened 2015-10-16
The good traceback is hidden by the useless traceback


micheles:
For instance, replace `category` with `assetCategory` in the exposure and you will get

```python
Traceback (most recent call last):
  File "/home/michele/oq-risklib/openquake/calculators/base.py", line 115, in run
    self.pre_execute()
  File "/home/michele/oq-risklib/openquake/calculators/event_based.py", line 388, in pre_execute
    super(EventBasedRuptureCalculator, self).pre_execute()
  File "/home/michele/oq-risklib/openquake/calculators/base.py", line 275, in pre_execute
    self.read_exposure_sitecol()
  File "/home/michele/oq-risklib/openquake/calculators/base.py", line 295, in read_exposure_sitecol
    self.exposure = readinput.get_exposure(self.oqparam)
  File "/home/michele/oq-risklib/openquake/commonlib/readinput.py", line 670, in get_exposure
    exposure, assets_node = get_exposure_lazy(fname)
  File "/home/michele/oq-risklib/openquake/commonlib/readinput.py", line 647, in get_exposure_lazy
    exposure['id'], exposure['category'],
  File "/home/michele/oq-risklib/openquake/commonlib/node.py", line 314, in __getitem__
    return self.attrib[i]
KeyError: 'category'
[2015-10-16 15:25:55,618 hazard job #429 - CRITICAL MainProcess/23734] Traceback (most recent call last):
  File "/home/michele/oq-engine/openquake/engine/engine.py", line 218, in run_calc
    _do_run_calc(calculator, exports)
  File "/home/michele/oq-engine/openquake/engine/engine.py", line 254, in _do_run_calc
    calc.pre_execute()
  File "/home/michele/oq-risklib/openquake/calculators/event_based.py", line 570, in pre_execute
    super(EventBasedCalculator, self).pre_execute()
  File "/home/michele/oq-risklib/openquake/calculators/base.py", line 267, in pre_execute
    self.csm = precalc.csm
  File "/home/michele/oq-risklib/openquake/commonlib/datastore.py", line 378, in getter
    value = self.datastore[key]
  File "/home/michele/oq-risklib/openquake/commonlib/datastore.py", line 283, in __getitem__
    raise KeyError('No %r found in %s' % (key, self))
KeyError: "No 'composite_source_model' found in <DataStore 429>"
```

Only the first part of the traceback is useful, the rest is noise.

PR #477, opened 2015-10-28
Removed a finally clause eating important errors

micheles:
Fixes https://github.com/gem/oq-risklib/issues/442. The same kind of problem also happened to one of our users with an invalid source model in an event based calculation.


ptormene:
LGTM

Issue #484, opened 2015-10-29
Add a test for the event based loss curves


micheles:
test_case_1 is the right test to extend. It should be done after we refactor the event based risk outputs (https://github.com/gem/oq-risklib/issues/475). We should also add a test for the loss maps.

PR #485, opened 2015-10-29
Fixed the loss curves saving

micheles:
There was a type bug so that the curves were not stored correctly in the HDF5 file. I am extending the EventBasedRisk demo to produce the curves. A test will be added later on (see https://github.com/gem/oq-risklib/issues/484).
The demos are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1155/console


ptormene:
As soon as demos complete successfully, LGTM

Issue #476, opened 2015-10-27
If an user pass a fragility function instead of a vulnerability function (or viceversa) a clear error should be raised


micheles:
This happened to Catalina. The error was:
 
```python
File "/home/michele/oq-risklib/openquake/commonlib/oqvalidation.py", line 156, in __init__
    self.risk_imtls = get_imtls(get_vfs(self.inputs))
  File "/home/michele/oq-risklib/openquake/commonlib/riskmodels.py", line 267, in get_imtls
    imls = list(rf.imls)
AttributeError: 'FragilityFunctionList' object has no attribute 'imls'
```

Issue #471, opened 2015-10-26
Unify the management of all risk models


micheles:
FragilityModels and ConsequenceModels are already unified, VulnerabilityModels should follow the same approach.

PR #486, opened 2015-10-29
Unify the management of the vulnerability model with the other risk models

micheles:
Closes https://github.com/gem/oq-risklib/issues/476 and https://github.com/gem/oq-risklib/issues/471. The demos are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1166


raoanirudh:
As far as the issue in #476 is concerned, LGTM.

Issue #480, opened 2015-10-28
If the exposure misses `deductible` and `insuredLimit` attributes you cannot compute insured losses


micheles:
The error should be raised early, before the computation.

raoanirudh:
The following is a valid use case:

* The exposure model defines structural, nonstructural, and contents costs for all assets. Insurance limits and deductibles are provided for the structural cost type, but not for nonstructural and contents.

* In the risk job file, `insured_losses = true`.

For this case, the insured losses should be computed for structural losses.

micheles:
If insuranceLimit and deductibles are missing for nonstructural and contents, make sure that the lines

nonstructural_vulnerability_file = nonstructural_vulnerability_model.xml
contents_vulnerability_file = contents_vulnerability_model.xml

are commented, so the calculation runs for structural.

Another idea (for the future) is to change insured_losses from a flag to a comma-separated string;
then we could define

insured_losses=structural

and have the insured losses to be computed only for structural, whereas for nonstructural and contents we could compute the losses without insurance.

PR #488, opened 2015-10-29
When computing insured losses for an exposure without deductible and insuredLimit, raise an error early

micheles:
Closes https://github.com/gem/oq-risklib/issues/480. The demos are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1159


ptormene:
LGTM

Issue #499, opened 2015-11-03
Make sure an agg_loss_curve output is generated


micheles:
This is needed for the Turkey computation. We need to produce an `agg_curve` for each realization. It must be possible to avoid computing the asset-specific losses (`avg_losses=False`), to reduce the data transfer and memory occupation.

PR #501, opened 2015-11-03
Building an aggregate loss curve per realization

micheles:
Closes https://github.com/gem/oq-risklib/issues/499. The demos are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1174/


openquake/calculators/event_based_risk.py
```diff
@@ -46,6 +82,17 @@ def square(L, R, factory):
     return losses
 
 
+def old_loss_curves(asset_values, rcurves, ordinal, ratios):
+    """
+    Build loss curves in the old format (i.e. (losses, poes)) from
+    loss curves in the new format (i.e. poes).
+    """
```
ptormene: You could add the parameters description to the docstring

micheles: Uhm... instead, I will add an underscore in front of this function, after all it is an internal detail that may even disappear in the future.

ptormene: ok

ptormene:
LGTM

PR #504, opened 2015-11-04
Update ScenarioDamage demo job.ini

CatalinaYepes:
* Change the job `description`
* Remove extra `export_dir` statement

### DIFF
--- a/demos/risk/ScenarioDamage/job.ini
+++ b/demos/risk/ScenarioDamage/job.ini
@@ -1,5 +1,5 @@
 [general]
-description = Scenario Hazard for Damage Demo (Nepal)
+description = Scenario Damage Demo (Nepal)
 calculation_mode = scenario_damage
 random_seed = 42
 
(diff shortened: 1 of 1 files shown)


demos/risk/ScenarioDamage/job.ini
```diff
@@ -15,9 +15,6 @@ asset_hazard_distance = 20
 [fragility]
 fragility_file = structural_fragility_model.xml
 
-[export]
-export_dir = /tmp/
-
```
micheles: Why do you want to remove the export_dir? The `export_dir` is a good thing, otherwise you will not know where the outputs will be exported when using the ``--exports`` option.

CatalinaYepes: because the file had repeated the `export_dir` (also in line 38)

micheles:
Ah, okay, then LGTM

PR #505, opened 2015-11-04
Update fault_rupture in ScenarioDamage demo

CatalinaYepes:
* The coordinates of the fault should be given as paris (lon, lat) in the `<gml:posList>`. To be consistent with the coordinates specified in the hypocenter, then the same convention should be followed.
* Some tabs were removed


micheles:
LGTM

Issue #510, opened 2015-11-05
Statistics for insured loss maps are not generated


micheles:
They were never generated by the engine. It looks like a missing functionality, check if we want them.

micheles:
Yes, we want them to be able to use a symmetric data structure. Actually they were generated by the engine, just not saved.

PR #513, opened 2015-11-06
Stored the event loss table and the full event loss table as single matrices

micheles:
HDF5 datasets can be compressed, so it is possible to store in a compact way matrices with a large number of zeros. That means that we can replace a nested structure loss_type -> realization (i.e. L x R distinct arrays of different lengths) with a single array of size (N,L, R, 2) where N is the number of assets and R the number of realizations. For the full loss table the approach of using a compressed matrix of shape (N, E, L, R, 2) does not work, since the saving time becomes huge. Instead we can use an extendable matrix with dtype (rup_id, ass_id, loss) where loss is a matrix of shape (L, R, 2) and apply the compression to that. With such an optimization it is actually possible to build the full loss table even for large computations like the Germany one: in that case 99.6% of the rows are zero, so the ~600 GB of expected size are reduced to a mere 2.4 GB of data transfer; on top of that, the compression reduce the disk space occupation to ~137 MB. Given such impressive results, I am removing the `specific_assets` feature and introducing an `asset_loss_table` flag. 

Closes https://github.com/gem/oq-risklib/issues/510.
Jenkins is happy: https://ci.openquake.org/job/zdevel_oq-risklib/1193
Anirudh's additional tests are running here: https://ci.openquake.org/job/zdevel_oq-risk-tests/73/console


openquake/calculators/event_based_risk.py
```diff
@@ -114,11 +112,13 @@ def event_based_risk(riskinputs, riskmodel, rlzs_assoc, assets_by_site,
...
     lti = riskmodel.lti  # loss type -> index
     L, R = len(lti), len(rlzs_assoc.realizations)
+    ela_dt = numpy.dtype([('rup_id', U32), ('ass_id', U32),
+                          ('loss', (F32, (L, R, 2)))])
 
     def zeroN2():
         return numpy.zeros((monitor.num_assets, 2))
     result = dict(AGGLOSS=square(L, R, list),
-                  SPECLOSS=square(L, R, list),
+                  SPECLOSS=[],
```
ptormene: Please rename into something like `ALL_LOSS`

openquake/calculators/event_based_risk.py
```diff
@@ -59,8 +61,7 @@ def build_agg_curve(lr_list, insured_losses, ses_ratio, curve_resolution,
         if len(data) == 0:  # realization with no losses
             continue
         for i in range(insured_losses + 1):  # insured_losses
```
ptormene: Please check if this can be substituted with `range(2)`

openquake/calculators/event_based_risk.py
```diff
@@ -131,19 +131,15 @@ def zeroN2():
...
 
-            # collect losses for specific assets
-            specific_ids = set(a.idx for a in out.assets
-                               if a.id in specific_assets)
-            if specific_ids:
+            if monitor.asset_loss_table:
                 for rup_id, all_losses, ins_losses in zip(
                         rup_ids, out.event_loss_per_asset,
                         out.insured_loss_per_asset):
                     for aid, sloss, iloss in zip(
```
ptormene: Please rename `sloss -> all_loss`

openquake/calculators/event_based_risk.py
```diff
@@ -297,88 +330,67 @@ def post_execute(self, result):
         :param result:
             a numpy array of shape (O, L, R) containing lists of arrays
         """
+        fullsize = numpy.prod(self.agg_losses.shape) * 4  # 32 bit floats
+        self.agg_losses.attrs['nonzero_fraction'] = (
+            self.agg_losses.attrs['nbytes'] / fullsize)
+
+        if self.oqparam.asset_loss_table:
+            dset = self.all_losses.dset
+            dset.attrs['nonzero_fraction'] = len(dset) / (self. N * self.E)
```
ptormene: Please remove the space in `self. N`

openquake/calculators/event_based_risk.py
micheles: No, it would do a needless computation of insured curves starting from NaN values if insured_losses is false.

openquake/calculators/event_based_risk.py
```diff
@@ -200,13 +209,12 @@ class EventBasedRiskCalculator(base.RiskCalculator):
...
 
     epsilon_matrix = datastore.persistent_attribute('epsilon_matrix')
-    spec_indices = datastore.persistent_attribute('spec_indices')
     is_stochastic = True
 
     outs = collections.OrderedDict(
         [('AVGLOSS', 'avg_losses-rlzs'),
-         ('AGGLOSS', 'agg_losses-rlzs'),
-         ('SPECLOSS', 'specific-losses-rlzs'),
+         ('AGGLOSS', 'agg_losses'),
```
raoanirudh: It might be better to call this `agg_loss_table` so as to use the same terminology as the `asset_loss_table`

openquake/commonlib/export/risk.py
```diff
@@ -44,6 +44,10 @@ def compose_arrays(a1, a2):
     all the fields. The two arrays must have the same length.
     """
     assert len(a1) == len(a2),  (len(a1), len(a2))
+    if a1.dtype.names is None and len(a1.shape) == 1:
+        # the first array is not composite, but its is one-dimensional
```
ptormene: Typo `its is`

openquake/calculators/event_based_risk.py
micheles: The name `agg_losses` is consistent with the name used in scenario_risk. Also, `agg_losses` is implemented as a simple array, whereas `asset_loss_table` is a sequence of records, i.e. a "table" in my mind. Should we change the implementation to use a sequence of records also for the `agg_losses` we could change the name. Notice that the name visible to the user after a run of the engine (the `display_name`) can be anything we want.

openquake/commonlib/export/risk.py
```diff
@@ -169,6 +203,50 @@ def export_avg_losses_stats(ekey, dstore):
     return fnames
 
 
+# this is used by classical_risk and event based_risk
```
ptormene: typo in the comment: `event based_risk -> event_based_risk`

openquake/commonlib/export/risk.py
```diff
@@ -169,6 +203,50 @@ def export_avg_losses_stats(ekey, dstore):
...
+    for rlz in rlzs:
+        losses = agg_losses[:, rlz.ordinal]
+        dest = dstore.export_path('agg_losses-rlz%03d.csv' % rlz.ordinal)
+        data = compose_arrays(tags, losses)
+        writers.write_csv(dest, data, fmt='%10.6E')
+        fnames.append(dest)
+    return fnames
+
+
+# this is used by event based_risk
```
ptormene: same as above

ptormene:
LGTM

openquake/commonlib/export/risk.py
micheles: Fixed

PR #514, opened 2015-11-06
Adds epistemic uncertainties on fault geometry into the Source Model Logic Tree

g-weatherill:
This PR adds support for a new family of epistemic uncertainties on the source model, specifically the ability to define different fault geometries for a given fault source (simple, complex and characteristic). 

The new epistemic uncertainties include:
i) `simpleFaultDipRelative` - increases or decreases a fault dip on a simple fault by a given amount

ii) `simpleFaultDipAbsolute` - replaces the dip on a simple fault with the specified value

iii) `simpleFaultGeometryAbsolute` - replaces the simple fault geometry with a different configuration

iv) `complexFaultGeometryAbsolute` - replaces the complex fault geometry with a different configuration

v) `characteristicFaultGeometryAbsolute` - replaces the characteristic fault geometry with a different configuration

These features accompany the changes in the oq-hazardlib introduced here: https://github.com/gem/oq-hazardlib/pull/388

Further modification was made to the `openquake.commonlib.logictree.BaseLogicTree` class to allow it to support LiteralNode instances in the `uncertaintyModel` node, rather than only text. This the epistemic uncertainties can be input as nrml SimpleFaultGeometry, ComplexFaultGeometry and CharacteristicFaultGeometry representations. With this functionality, also modified here is the logic tree representations of `incrementalMFDAbsolute` (introduced here https://github.com/gem/oq-risklib/pull/467) to return the incremental MFD representation to its full nrml format rather than the arbitrary string format adopted previously.

Tests and demos are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1205


openquake/commonlib/logictree.py
```diff
@@ -379,8 +414,35 @@ def apply_uncertainty(self, value, source):
...
             # source didn't pass the filter
             return
-
-        self._apply_uncertainty_to_mfd(source.mfd, value)
+        if self.uncertainty_type in MFD_UNCERTAINTY_TYPES:
+            self._apply_uncertainty_to_mfd(source.mfd, value)
+        elif self.uncertainty_type in GEOMETRY_UNCERTAINTY_TYPES:
+            self._apply_uncertainty_to_geometry(source, value)
+        else:
+            pass
```
mmpagani: Why 'pass' instead of generating an error?

mmpagani:
I haven't dowloaded the code locally. Would like to do it tomorrow. So far great job Graeme!

openquake/commonlib/logictree.py
g-weatherill: It should be an error (I will add it)

micheles:
I have two general comments. The first is that I would like to include an end-to-end test exercising the feature, or a modification of one the existing demos using the feature. The second is a design suggestion. Having a method with a long string of `ifs`, like this

```python
class Parser(object):

    def parse_uncertainty_value(self, node, branchset):
        if branchset.uncertainty_type == 'sourceModel':
            return 1
        elif branchset.uncertainty_type == 'abGRAbsolute':
            return 2
        elif branchset.uncertainty_type == 'incrementalMFDAbsolute':
            return 3
        elif branchset.uncertainty_type == 'simpleFaultGeometryAbsolute':
            return 4
        elif branchset.uncertainty_type == 'complexFaultGeometryAbsolute':
            return 5
        elif branchset.uncertainty_type == 'characteristicFaultGeometryAbsolute':
            return 6
        else:
            return 0
```
is ugly. Everywhere in risklib we use a different pattern (see for occurrences of CallableDict in the calculators, the distribution functions, the exporters and even the nrml parser) which is a dictionary-based dispatcher. Every branch of the `ifs` becomes an independent function and all functions are collected together by a decorator. The code before would become something like this:

```python
from openquake.baselib.general import CallableDict

parse_uncertainty_value = CallableDict(
    keyfunc=lambda branchset: branchset.uncertainty_type,
    keymissing=lambda branchset, node: 0)

@parse_uncertainty_value.add('sourceModel')
def parse_sourceModel(branchset, node):
    return 1

@parse_uncertainty_value.add('abGRAbsolute')
def parse_abGRAbsolute(branchset, node):
    return 2

@parse_uncertainty_value.add('incrementalMFDAbsolute')
def parse_incrementalMFDAbsolute(branchset, node):
    return 3

@parse_uncertainty_value.add('simpleFaultGeometryAbsolute')
def parse_simpleFaultGeometryAbsolute(branchset, node):
    return 4

@parse_uncertainty_value.add('complexFaultGeometryAbsolute')
def parse_complexFaultGeometryAbsolute(branchset, node):
    return 5

class Branchset:
    uncertainty_type = 'sourceModel'

class Node:
    pass

print parse_uncertainty_value(Branchset(), Node())
```

The CallableDict approach has the advantage of being easily extensible and uniformity with the rest of the code.

g-weatherill:
Two end-to-end tests are now added. Refactoring the design is postponed for a future PR

openquake/commonlib/logictree.py
```diff
@@ -797,27 +934,117 @@ def validate_uncertainty_value(self, node, branchset, value):
...
-                if _float_re.match(min_mag) and _float_re.match(bin_width) and\
-                        len(rates):
-                    return
+            min_mag, bin_width = (node.incrementalMFD["minMag"],
+                                  node.incrementalMFD["binWidth"])
+            rates = node.incrementalMFD.occurRates.text.strip()
+            try:
+                rates = valid.positivefloats(rates)
+            except ValueError:
+                rates = []
```
micheles: Are we hiding the error here? If an user inserts wrong rates instead of getting an error the rates mysteriously
disappear?

openquake/commonlib/logictree.py
```diff
@@ -797,27 +934,117 @@ def validate_uncertainty_value(self, node, branchset, value):
...
+                         node.lowerSeismoDepth.text.strip(),
+                         node.dip.text.strip())
+        spacing = node["spacing"].strip()
+        try:
+            # Parse the geometry
+            coords = split_coords_2d(map(float,
+                                         node.LineString.posList.text.split()))
+            trace = geo.Line([geo.Point(*p) for p in coords])
+        except ValueError:
+            trace = []
```
micheles: Same as before. The trace disappear while I would have expected a ValidationError to be raised. Am I missing something?

openquake/commonlib/logictree.py
```diff
@@ -797,27 +934,117 @@ def validate_uncertainty_value(self, node, branchset, value):
...
+        """
+        valid_edges = []
+        for edge_node in node.nodes:
+            try:
+                coords = split_coords_3d(map(
+                    float,
+                    edge_node.LineString.posList.text.split()))
+                edge = geo.Line([geo.Point(*p) for p in coords])
+            except ValueError:
+                edge = []
```
micheles: Same.

micheles:
I have generated the .rst reports for the tests 20 and 21, for consistency with the other tests and also to see better the logic tree.

openquake/commonlib/logictree.py
g-weatherill: Well, I was trying to avoid raising the ValueError directly as it may not be particularly useful to the user, and instead just rely on the later validation error so that it points them to the node that fails. So by using the try-except to catch the value error and, if raised, turn the trace into a simpler boolean that can be tested easily in the following line (i.e. if the trace was valid then len(trace) != 0 and is hence True, otherwise False), then we ensure that only the ValidationError is raised.

micheles: Then please add a comment explaining the validation logic.

openquake/commonlib/logictree.py
micheles: The node library has a context manager to manage the errors correctly, including the line number. The trick is to write

```python
with context(self.filename, node):
    rates = valid.positivefloats(rates)
```

If the rates are invalid you will get a nice error including the line number. Please give it a try.

micheles:
LGTM

mmpagani:
LGTM

Issue #515, opened 2015-11-08
It would be useful to have a show_attrs command


micheles:
To show the attributes of a HDF5 dataset. For instance, after running an event based risk calculation one could give the command

```bash
$ oq-lite show_attrs -1 agg_losses
```

to see the value of the `nonzero_factor` attribute.

PR #516, opened 2015-11-09
Added show_attrs command

micheles:
Closes https://github.com/gem/oq-risklib/issues/515.


openquake/commonlib/tests/commands_test.py
```diff
@@ -102,6 +106,11 @@ def test_show_calc(self):
 lon:float64:,lat:float64:
 0.00000000E+00,0.00000000E+00''')
 
+    def test_show_attrs(self):
```
ptormene: Please also add 2 more tests:
* key not found
* key has attributes

openquake/commonlib/commands/show_attrs.py
```diff
@@ -0,0 +1,43 @@
...
+#  GNU General Public License for more details.
+
+#  You should have received a copy of the GNU Affero General Public License
+#  along with OpenQuake.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import print_function
+from openquake.commonlib import sap, datastore
+
+
+def show_attrs(calc_id, key):
```
ptormene: Please consider if it might make sense to show all attributes if the `key` parameter is not given

openquake/commonlib/tests/commands_test.py
micheles: Added

openquake/commonlib/commands/show_attrs.py
micheles: I do not see this as a needed feature for the moment. Usually one is only interested to a specific kind of outputs.

ptormene:
LGTM

PR #517, opened 2015-11-09
Added a validation on the cost types

micheles:
The business_interruption cost_type gave in error in Anirudh's tests (https://ci.openquake.org/job/zdevel_oq-risk-tests/73/console) since it is 21 characters long and the limit was set to 20, so it was silently truncated.

### DIFF
--- a/openquake/commonlib/readinput.py
+++ b/openquake/commonlib/readinput.py
@@ -638,8 +638,10 @@ class DuplicatedID(Exception):
     Raised when two assets with the same ID are found in an exposure model
     """
 
-cost_type_dt = numpy.dtype([('name', '|S20'), ('type', '|S20'),
-                            ('unit', '|S20')])
+COST_TYPE_SIZE = 21  # using 21 chars since business_interruption has 21 chars
+cost_type_dt = numpy.dtype([('name', (bytes, COST_TYPE_SIZE)),
+                            ('type', (bytes, COST_TYPE_SIZE)),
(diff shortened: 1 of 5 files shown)


openquake/commonlib/readinput.py
```diff
@@ -670,12 +672,20 @@ def get_exposure_lazy(fname, ok_cost_types):
...
     cost_types = [(ct['name'], ct['type'], ct['unit'])
                   for ct in conversions.costTypes
                   if ct['name'] in ok_cost_types]
     if 'occupants' in ok_cost_types:
         cost_types.append(('occupants', 'per_area', 'people'))
     cost_types.sort(key=operator.itemgetter(0))
+    for row in cost_types:
+        valid.simple_id(row[0])  # the name
+        for col in row:
+            if len(col) > COST_TYPE_SIZE:
```
raoanirudh: Presuming that we have already checked for the validity of the strings in cost_types, would we ever reach this point?

micheles: You are right that the test should be done earlier. Let me see if I can move this up in the validation chain (right now there is no check at all).

micheles:
Move the check earlier; now only structural,nonstructural,contents and business_interruption are accepted as cost types.

raoanirudh:
Are occupants handled separately?

micheles:
I am not happy about how occupants are managed (they should be more like cost types) but let leave them out of this pull request.

raoanirudh:
LGTM

Issue #503, opened 2015-11-03
Make sure the XML exporters work for loss curves and maps


micheles:
Including the agg_curve.

PS: Anirudh discovered tha the loss map exporter fails when there are occupants.

PR #520, opened 2015-11-11
Loss map exporter

micheles:
I was incorrectly exporting the avglosses as loss maps. The EventBasedRisk demo is using it:
https://ci.openquake.org/job/zdevel_oq-risklib/1194/console

See https://github.com/gem/oq-risklib/issues/503


raoanirudh:
LGTM

PR #532, opened 2015-11-17
Fixed an export bug in the agg_losses

micheles:
The tests are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1209/

To see the difference, compare https://github.com/gem/oq-risklib/blob/master/openquake/qa_tests_data/event_based_risk/case_1/expected/agg_losses-rlz000.csv
with https://github.com/gem/oq-risklib/blob/agg-losses-export/openquake/qa_tests_data/event_based_risk/case_1/expected/agg_losses-rlz000.csv


daniviga:
LGTM if tests are green

Issue #482, opened 2015-10-28
Remove the epsilon_sampling "feature"


micheles:
Now the system should be fast enough to make it possible to remove that optimization. For very large computations the data transfer/memory occupation could grow too much. In that case we should ask the user to set `asset_correlation=0` and to use a different strategy for that situation, when we don't need to build the correlation matrix. In that case the epsilons can be built on the workers by using very little memory, provided we use a seed for each rupture, so that there is no block size dependence.

Issue #372, opened 2015-09-14
Add a test of block size independence, even in situations with an epsilon_sampling


micheles:
This happens in the event based risk calculator. We can just check the total average loss.

PR #533, opened 2015-11-17
Removed epsilon sampling from event_based_risk

micheles:
Closes https://github.com/gem/oq-risklib/issues/482. Also closes https://github.com/gem/oq-risklib/issues/372, since it will become meaningless. With this change the data transfer in the demos is reduced from ~609 MB to ~125 MB, see https://ci.openquake.org/job/zdevel_oq-risklib/1208/console.


openquake/risklib/riskinput.py
```diff
@@ -432,13 +394,11 @@ def compute_expand_gmfs(self):
             gmfa[i] = expanded_gmf
         return gmfa  # array R x N
 
-    def get_all(self, rlzs_assoc, assets_by_site, eps):
+    def get_all(self, rlzs_assoc, assets_by_site):
         """
```
ptormene: Missing params description

micheles: Added

ptormene:
LGTM

Issue #534, opened 2015-11-17
Save an asset_loss_table and an agg_loss_table for each realization


micheles:
This was requested by Vitor. Currently there is a single big matrix for the asset_loss_table and a single big matrix for the agg_loss_table.

PR #535, opened 2015-11-17
Use one loss table per realization

micheles:
Closes https://github.com/gem/oq-risklib/issues/534. The demos are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1211/
Now the `agg_loss_table` has a structure similar to the `asset_loss_table`.
Testing with the Germany model shows a reduction in the data transfer and size of the asset_loss_table from 2.42 GB to 1.16 GB, thanks to the removal of many zeros in the loss matrices. Turkey sees a similar reduction from 100 GB to 58 GB.

NB: in test_case_1 I reduced ses_per_logic_tree_path from 2,000 to 20, for easy of debugging.


ptormene:
LGTM

PR #537, opened 2015-11-17
Removed a spurios comma from SingleRuptureSource.source_id

micheles:
This is breaking one CSV reader in the RMTK, as reported by Catalina and Anirudh.


CatalinaYepes:
LGTM

Issue #541, opened 2015-11-19
Removing `__slots__` in hazardlib


micheles:
I have wanted to remove the `__slots__` in hazardlib for three years, because I always suspected that they were useless. Today I have the proof of that: I ran the Turkey calculation (which is a particularly memory hungry calculation using the SHARE model) with and without slots and there is no measurable difference in the memory occupation. Removing the `__slots__` means avoid a large
class of pickling errors, such as forgetting to add `_mesh` to the slots of PlanarRupture and running into the [problem submitted by Blaž Barič](https://groups.google.com/d/msg/openquake-users/D8vf1AXNXDM/Zj-ECHmbBgAJ).

PR #542, opened 2015-11-19
Removed with_slots

micheles:
Companion of https://github.com/gem/oq-hazardlib/pull/397. This is required for https://github.com/gem/oq-risklib/issues/541. In the future we may or may not remove the `__slots__` from the Node class, depending on a memory performance analysis.
The tests are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1218


g-weatherill:
LGTM

Issue #543, opened 2015-11-19
The hazard maps are exported by using the hazard curve exporter


micheles:
Instead we should use the hazard map exporter! We need to add a test on that.

PR #544, opened 2015-11-19
Fix hazard map exporter

micheles:
Fix a bug discovered by Catalina, see https://github.com/gem/oq-risklib/issues/543, and backported the fix https://github.com/gem/oq-risklib/pull/537


micheles:
Manually tested that all tests are passing, merging in 1.6 and running the tests again.

PR #549, opened 2015-11-24
Fixed hazard_curve_name

micheles:
I was using a wrong regex. Then right splitting of `rlz-000` is `('000', '')` and not `('00', '0')`.
The tests are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1226/


daniviga:
LGTM

Issue #429, opened 2015-10-13
Unskip the tests case_1 and case_2 in classical risk


micheles:
It means cleaning up the "load from hazard curves" functionality. It should work like the "load from gmf" in scenario_risk and scenario_damage.

PR #550, opened 2015-11-25
Ported the --load-curve functionality from the engine to oq-lite

micheles:
This is the first step towards https://github.com/gem/oq-risklib/issues/429


ptormene:
LGTM

Issue #521, opened 2015-11-11
We need a test for XML loss map export in classical_risk


micheles:
This is is missing and it is the source of the confusion with event based loss map exporter.
Idem for loss curves, unless we deprecate the XML format.

PR #553, opened 2015-11-26
Building and exporting loss maps from classical_risk

micheles:
Closes https://github.com/gem/oq-risklib/issues/521


ptormene:
LGTM

PR #554, opened 2015-11-26
Back to using nosetests in Travis

micheles:
Now we can use the options --with-doctest and --with-xunit.
Closes https://github.com/gem/oq-risklib/issues/489.


daniviga:
If it's fine to skip the slow tests, then LGTM

PR #559, opened 2015-12-01
Reduced the exposure in classical_risk/test_case_3

micheles:
To make the test faster. Also, I have regenerated the .rst reports.


ptormene:
LGTM

PR #560, opened 2015-12-01
Fixed export bug

micheles:
This was discovered by Catalina. The sort in get_assets was different from the sort in build_assetcol in the case of multiple assets on the same point.


daniviga:
LGTM if tests are green and @CatalinaYepes happy

CatalinaYepes:
LGTM

PR #565, opened 2015-12-02
Added a check on the uniform_hazard_spectra parameter

micheles:
The problem was discovered by an user in Taiwan, trying to compute UHS curves for PGV. The error should happen at the beginning, not at the end of the computation.


debian/changelog
```diff
@@ -1,4 +1,5 @@
   [Michele Simionato]
+  * Added a check on the uniform_hazard_specta parameter depending on the IMTs
```
ptormene: Typo: spectRa

ptormene:
LGTM

PR #569, opened 2015-12-02
Print pstat info directly when using the --profile option

micheles:
This is convenient if you don't have runsnake installed. For instance

```bash
$ oq-lite run job.ini -c0 --profile 20
```

prints the 20 slowests operations performed by oq-lite.


openquake/commonlib/commands/run.py
```diff
@@ -16,15 +16,45 @@
...
+    :param pstatfile: path to a .pstat file
+    :param n: the maximum number of stats to retrieve
+    """
+    stream = io.BytesIO()
+    ps = pstats.Stats(pstatfile, stream=stream)
+    ps.sort_stats('cumtime')
+    ps.print_stats(n)
+    lines = stream.getvalue().splitlines()
+    for i, line in enumerate(lines):
+        if line.startswith('   ncalls'):
```
ptormene: It would be more readable if you could add a small comment with an example of what you expect to read (useful to understand here and line 51)

ptormene:
Besides a minor comment, LGTM

openquake/commonlib/commands/run.py
micheles: Done

ptormene: Actually I meant an example of what you read in `line`, but also the example of output looks good

PR #572, opened 2015-12-03
Avoid computing the rjb distance twice

micheles:
This PR takes advantage of the new filtering in the hazardlib ContextMaker (https://github.com/gem/oq-hazardlib/pull/400) to avoid computing the rjb distances twice. As an use case, I considered our demo
LogicTreeCase2ClassicalPSHA on gemsun02. Then the operation "getting ruptures" (which before involved filtering) passed from 403s to 231s. Still the net effect on the total runtime is negligible.
The tests and demos are green: https://ci.openquake.org/job/zdevel_oq-risklib/1243/


openquake/commonlib/commands/run.py
```diff
@@ -93,14 +123,16 @@ def run(job_ini, concurrent_tasks=None, pdb=None,
     :param exports:
         export type, can be '', 'csv', 'xml', 'geojson' or combinations
     :param profile:
-        when True, enable Python cProfile functionality
+        enable Python cProfile functionality
```
ptormene: you might explicitly declare the type like
`:type profile: bool`

ptormene:
Please check about travis failing. Apart from that, LGTM

Issue #566, opened 2015-12-02
Port the classical_damage calculator to oq-lite


micheles:
With all of its tests.

PR #573, opened 2015-12-03
Porting the classical damage calculator to oq-lite

micheles:
Closes https://github.com/gem/oq-risklib/issues/566. Tests and demos are running: https://ci.openquake.org/job/zdevel_oq-risklib/1249.
Notice that the CSV exporter for the damages has been unified with the other exporter, so the expected CSV files have changed also for other calculators (now they display asset_ref, lon, lat and the ordering is first by lon, lat and then by asset_ref).


ptormene:
LGTM

PR #576, opened 2015-12-04
Improved an error message

micheles:
Now if a task tries to call `monitor.flush()` in the error message we get the name of the task and the name of the monitor. The tests are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1248/

NB: this requires https://github.com/gem/oq-hazardlib/pull/401


micheles:
We broke master having merged the change in hazardlib without merging this PR: https://ci.openquake.org/job/master_oq-risklib/1759/

daniviga:
LGTM if tests are green

PR #580, opened 2015-12-14
Fixed minor bug with steps_per_interval=1

micheles:
An `if` was wrong, as discovered by Anirudh, so there was an
`AttributeError: 'FragilityFunctionList' object has no attribute 'interp_imls'` for `steps_per_interval=1`.


raoanirudh:
LGTM

PR #583, opened 2015-12-15
Small fixes to the monitoring mechanism

micheles:
I am calling monitor.flush() in apply_reduce even in the case of no_distribute, so there is no need to flush in client code. This avoids common errors resulting in missing performance information. Also, I am changing the classical calculator to not return the monitor, since it is already returned implicitly
by the litetask decorator.


ptormene:
LGTM

PR #590, opened 2015-12-16
Added an export for the realization table

micheles:
This was requested by Catalina. Now the associations between realization number and logic tree path is available in a file called `realizations.csv`. We also report the weight of the realization.


CatalinaYepes:
LGTM

PR #591, opened 2015-12-17
Export empty files too

micheles:
Before empty files were not exported: this cause an issue with the engine, which was trying to zip non-exiting files.

PS: while working at it, I discovered and fixed a small bug with invalid .hdf5 files affecting the show command.


daniviga:
LGTM

Issue #588, opened 2015-12-16
Sometime report.rst cannot be written to /tmp


daniviga:
By default the `report.rst` file generated by `oq-lite info -r` is saved to `/tmp`; this can lead to an error if that file already exists but is owned by a different user.

There are at least 2 possible ways to fix it:
1. Make `oq-lite info -r` aware of the `output_dir` config in the `.ini` and then save the file in that folder (this reduces the clashes with other users)
2. Add the calculation id in the file name (like `report_xyz.rst`)

Reported by @CatalinaYepes

Issue #589, opened 2015-12-16
oq-lite '-f' flag does not work anymore


daniviga:
The oq-lite `-f` does not work anymore an gives an error when used:

```python
Traceback (most recent call last):
  File "/usr/local/openquake/oq-risklib/bin/oq-lite", line 36, in <module>
    oq_lite()
  File "/usr/local/openquake/oq-risklib/bin/oq-lite", line 33, in oq_lite
    parser.callfunc()
  File "/usr/local/openquake/oq-risklib/openquake/commonlib/sap.py", line 167, in callfunc
    return self.func(**vars(namespace))
  File "/usr/local/openquake/oq-risklib/openquake/commonlib/sap.py", line 201, in main
    return func(**kw)
  File "/usr/local/openquake/oq-risklib/openquake/commonlib/commands/info.py", line 115, in info
    _info(name, filtersources, weightsources)
  File "/usr/local/openquake/oq-risklib/openquake/commonlib/commands/info.py", line 96, in _info
    dic = groupby(assetcol, operator.attrgetter('taxonomy'))
  File "/usr/local/openquake/oq-hazardlib/openquake/baselib/general.py", line 622, in groupby
    kgroups = itertools.groupby(sorted(objects, key=key), key)
AttributeError: 'numpy.void' object has no attribute 'taxonomy'
```
This flag is also documented here: https://github.com/gem/oq-risklib/blob/master/doc/effective-realizations.rst so, if we want to remove it the documentation must be updated.

Reported by @CatalinaYepes

PR #592, opened 2015-12-17
Improved the report generation mechanism

micheles:
Closes the issues https://github.com/gem/oq-risklib/issues/588 and https://github.com/gem/oq-risklib/issues/589. While at it, I regenerated all of the reports.


daniviga:
LGTM

Issue #587, opened 2015-12-15
Make sure that the average losses are exported correctly


micheles:
Anirudh has a case in  https://github.com/gem/oq-risk-tests/tree/master-cases/test/event_based_risk/inputs/case_master where insured losses for different loss types seem to be swapped.

PR #593, opened 2015-12-17
Fixed the management of avg_losses in the event_based_risk calculator

micheles:
This was reported by Catalina. In the case of more than one realization and more of one loss type the numbers were all mixed up. This was never discovered before since avg_losses were a non-official feature which was previously tested in simple cases and then disabled by default. Also, according to Catalina's suggestion, I am changing the array structure to be consistent with the agg_losses.

Fixes https://github.com/gem/oq-risklib/issues/587


ptormene:
LGTM

PR #595, opened 2015-12-17
Avoided computing the rjb distance twice

micheles:
While at it, I am also reducing the numerical precision to 6 digits, to avoid random failures like here: https://ci.openquake.org/job/master_oq-risklib/1786/
The tests pass: https://ci.openquake.org/job/zdevel_oq-risklib/1264


ptormene:
LGTM

PR #600, opened 2015-12-21
Print a helpful message when killing a calculation with CTRL-C

micheles:
This is a workaround against a shortcoming of multiprocessing, that does not kill properly the processes and makes them children of `init`.


daniviga:
LGTM

PR #606, opened 2015-12-22
Several improvements to the classical and classical_tiling calculators

micheles:
Lots of small stuff. The ground breaking change was this commit in hazardlib, that saved a LOT of memory: https://github.com/gem/oq-hazardlib/commit/2a608d96cbe5fc6ab382de4766032908e60b4aa6


openquake/calculators/classical.py
```diff
@@ -413,43 +424,51 @@ class ClassicalTilingCalculator(ClassicalCalculator):
...
         sources = self.csm.get_sources()
-        for tile in tiles:
+        rlzs_assoc = self.csm.get_rlzs_assoc()
+        siteidx = 0
+        tmanagers = []
+        maximum_distance = self.oqparam.maximum_distance
+        num_blocks = math.ceil(self.oqparam.concurrent_tasks / len(tiles))
+
+        for i, tile in enumerate(tiles, 1):
+            monitor = self.monitor.new('tile')
```
ptormene: Is it correct to pass `'tile'` instead of `tile`?

micheles: The name is not relevant here. I am removing it.

ptormene:
LGTM

Issue #609, opened 2016-01-05
The source model logic tree file has to be in format NRML 0.4


micheles:
There should be no such dependency on the version, since there are no changes, in particular we should be able to use NRML 0.5. This was reported by Anirudh.

PR #610, opened 2016-01-05
Fix the source model logic tree parser to support NRML 0.5

micheles:
Closes https://github.com/gem/oq-risklib/issues/609


raoanirudh:
LGTM

PR #614, opened 2016-01-07
Revolutioned the source management mechanism

micheles:
The recent SARA computation highlighted the known shortcomings of the current
hazard calculators, thus making necessary a substantial change in the way
sources are managed. This pull requests implements the required changes.

1. the function `get_composite_model` has been changed. Now it returns the
full composite source model of the computation, without filtering the
sources and without splitting them. This has several advantages:

  a. the function is now much faster and much simpler; its API is just
     `get_composite_source_model(oqparam, in_memory=True)`
  b. the composite source model stored in the datastore is now much smaller
     in size, since unsplit sources take a LOT less space than split sources
  c. the stored composite source model is the real model, not an artificially
     split version of it: that makes several things (including debugging) easier

  The weighting of the sources has been kept, but is disabled when
  `in_memory=False`, which is used for fast analysis of the source model.

2. The filtering and splitting of the sources has been moved in a class
called `SorceManager`: all of the old `SourceProcessor` classes have been removed.

  a. the filtering is now done after the` CompositeSourceModel` has been instantiated (i.e. there is a better separation of concerns)
  b. the splitting is still done only for the filtered sources, but now it is done only for the heavy sources; the reduction in data transfer can be massive (i.e. from 43 GB to 750 MB, seen in the SARA model)
  c. a source is considered heavy if its weight is bigger that the `maxweight` attribute of the `CompositeSourceModel`; such value is computed by dividing the total weight of the source model by the `concurrent_tasks` parameter (or 1 if `concurrent_tasks=0`), thus we still try to generate a number of tasks of the order of `concurrent_tasks` even if it can be much larger if there are very heavy sources;
  d. the light sources are submitted unsplit before the heavy sources, so the computation is faster because it starts parallelizing as early as possible

3. Since the classical calculator is completely different, it was possible to hugely simplify the `classical_tiling` calculator and to make it distribute much better than before, when the sources were not split. 
 
4. A lot of refactoring has been performed. In particular now all the information about the sources, including source_id, tectonic region id, source class, filtering time, splitting time, calculation time, and
number of splits is saved in the datastore in a single place.

5. As a consequence of the change in the splitting mechanism, the number of sources generated and therefore the seeds generated,  there is onne seed for each (split) source, depend on the parameter
concurrent_tasks. Therefore forevent based calculations the numbers are expected to be completely different for quantities sensitive to the seeds. On the other hand, robust quantities stay robust and none of the original event based hazard test has been broken.

6. The numbers in the classical calculators have changed slightly. Doing the same computation with a split or an unsplit source in theory should give the same result, in practice gives results that differs slightly due to numerical errors, which have nothing to do with the seeds.

The demos are running: https://ci.openquake.org/job/zdevel_oq-risklib/1319/

This PR is blocked by https://github.com/gem/oq-engine/pull/1921 that must enter first. We need to remove the old calculators, otherwise the engine tests will break, since the numbers are different from the past.


openquake/commonlib/readinput.py
```diff
@@ -472,46 +472,41 @@ def get_source_models(oqparam, source_model_lt, sitecol=None, in_memory=True):
...
-    :param monitor:
-        a monitor instance
-    :param no_distribute:
-        used to disable parallel splitting of the sources
-    :param dstore:
-        a DataStore instance (possibly fake)
-    :returns:
-        an iterator over :class:`openquake.commonlib.source.SourceModel`
+    :param in_memory:
+        if False, just parse the XML without instanting the sources
```
ptormene: typo `instantiating`

openquake/commonlib/readinput.py
```diff
@@ -472,46 +472,41 @@ def get_source_models(oqparam, source_model_lt, sitecol=None, in_memory=True):
...
     for source_model in get_source_models(
-            oqparam, source_model_lt, processor.sitecol,
-            in_memory=hasattr(processor, 'process')):
+            oqparam, source_model_lt, in_memory=in_memory):
         for trt_model in source_model.trt_models:
+            trt_model.sources = sorted(trt_model, key=getid)
             trt_model.id = trt_id
+            for src in trt_model:
+                if hasattr(src, 'trt_model_id'):
+                    # .trt_model_id is missing for source nodes
```
ptormene: Please further explain why in some cases we need to enumerate, and in others we don't

openquake/commonlib/source.py
```diff
@@ -712,223 +707,144 @@ def collect_source_model_paths(smlt):
...
         self.monitor = monitor
+        self.rlzs_assoc = csm.get_rlzs_assoc()
+        self.split_map = {}
+        self.infos = {}  # trt_model_id, source_id -> SourceInfo tuple
+        logging.info('instantiating SourceManager with maxweight=%.1f',
+                     self.maxweight)
+
+    def get_sources(self, kind, sitecol):
+        """
+        Get and the sources of kind `kind` affecting the `sitecol`; split
```
ptormene: Please fix the docstring

openquake/qa_tests_data/event_based/case_7/expected/hazard_curve-mean.csv
```diff
@@ -1,4 +1,4 @@
...
+0.00000 1.00000,8.1712654E-01 7.8995929E-01 7.5859713E-01 7.2065396E-01 6.7792359E-01 6.1310979E-01 5.0606272E-01 3.6841306E-01 2.4286609E-01 1.5123441E-01 7.8646755E-02 3.4762479E-02 1.6618300E-02 7.2526026E-03 3.0707836E-03 1.3878477E-03 0.0000000E+00 0.0000000E+00 0.0000000E+00,8.5031363E-01 8.3037928E-01 8.0714866E-01 7.8087136E-01 7.4963580E-01 7.1754637E-01 6.7747697E-01 6.1787004E-01 5.2001434E-01 3.9718247E-01 2.7310472E-01 1.7546872E-01 1.0138969E-01 6.2703556E-02 3.0531054E-02 1.3713041E-02 5.4034178E-03 1.7969143E-03 7.0956520E-04
+0.00000 2.50000,6.0986540E-01 5.1536851E-01 3.8562112E-01 2.3938177E-01 1.1598324E-01 4.8509312E-02 1.5884493E-02 4.2072912E-03 1.6475275E-03 0.0000000E+00 0.0000000E+00 0.0000000E+00 0.0000000E+00 0.0000000E+00 0.0000000E+00 0.0000000E+00 0.0000000E+00 0.0000000E+00 0.0000000E+00,6.8524997E-01 6.4816312E-01 5.8752199E-01 4.8521846E-01 3.5791290E-01 2.2090211E-01 1.0904116E-01 4.3031374E-02 1.1852730E-02 2.0553484E-03 6.2968510E-04 6.2968510E-04 5.4972509E-04 0.0000000E+00 0.0000000E+00 0.0000000E+00 0.0000000E+00 0.0000000E+00 0.0000000E+00
```
ptormene: It looks like the first columns of the two files are more similar to each other, with respect to the final columns. Is this just a coincidence, or is it somehow expected?

openquake/commonlib/readinput.py
micheles: Fixed

openquake/commonlib/readinput.py
micheles: Done

openquake/commonlib/source.py
micheles: Done

openquake/commonlib/source.py
```diff
@@ -567,43 +558,47 @@ def trt_models(self):
...
-                for src in trt_model:
-                    nr = src.count_ruptures()
-                    src.weight = get_weight(src, num_ruptures=nr)
-                    num_ruptures += nr
-                trt_model.num_ruptures = num_ruptures
-                logging.info('Processed %s', trt_model)
+            weight = 0
+            num_ruptures = 0
+            for src in trt_model:
+                if src.__class__.__name__ in ('PointSource', 'AreaSource'):
```
ptormene: Please check if it's possible to change this using `isinstance`

openquake/commonlib/sourceconverter.py
```diff
@@ -34,6 +34,20 @@
...
+
+def get_weight(src, num_ruptures=None):
+    """
+    :param src: a hazardlib source object
+    :param num_ruptures: if None it is recomputed
+    :returns: the weight of the given source
+    """
+    num_ruptures = num_ruptures or src.count_ruptures()
+    weight = (num_ruptures * POINT_SOURCE_WEIGHT
+              if src.__class__.__name__ in ('PointSource', 'AreaSource')
```
ptormene: same as above

openquake/calculators/event_based.py
```diff
@@ -268,14 +267,16 @@ def __lt__(self, other):
...
     :param sources:
         List of commonlib.source.Source tuples
     :param sitecol:
         a :class:`openquake.hazardlib.site.SiteCollection` instance
-    :param info:
-        a :class:`openquake.commonlib.source.CompositionInfo` instance
+    :param rlzs_assoc:
+        a :class:`openquake.commonlib.source.RlzsAssoc` instance
+    :param siteidx:
+        always equal to 0
```
ptormene: Do we need any sanity check for this?

openquake/commonlib/reportwriter.py
```diff
@@ -93,9 +100,14 @@ def build_report(job_ini, output_dir=None):
     oq = readinput.get_oqparam(job_ini)
     output_dir = output_dir or os.path.dirname(job_ini)
     calc = base.calculators(oq)
+    calc.SourceManager = source.DummySourceManager
+    calc.is_effective_trt_model = lambda result_dict, trt_model: True
```
ptormene: Please add a comment about the fact we are not actually running the calculation, or use a dummy calculator to make it cleaner

openquake/calculators/event_based.py
micheles: Yes, I will add something

openquake/calculators/event_based.py
```diff
@@ -380,50 +386,35 @@ class EventBasedRuptureCalculator(base.HazardCalculator):
...
 
-        store_source_chunks(self.datastore)
-        logging.info('Generated %d SESRuptures',
-                     sum(len(v) for v in ruptures_by_trt.values()))
-
-        self.rlzs_assoc = self.csm.get_rlzs_assoc(
-            lambda trt: len(ruptures_by_trt.get(trt.id, [])))
-
-        return ruptures_by_trt
+    def agg_curves(self, acc, val):
```
ptormene: Please add a docstring

openquake/calculators/event_based.py
```diff
@@ -658,7 +649,7 @@ def post_execute(self, result):
             # use a different datastore
             self.cl = ClassicalCalculator(oq, self.monitor)
             self.cl.datastore.parent = self.datastore
-            result = self.cl.run(pre_execute=False)
+            result = self.cl.run()
```
ptormene: Please add a comment

openquake/commonlib/commands/info.py
```diff
@@ -78,17 +75,12 @@ def _info(name, filtersources, weightsources):
             sitecol = None
         if 'source_model_logic_tree' in oqparam.inputs:
             print('Reading the source model...')
-            if weightsources:
-                sp = source.SourceFilterWeighter
-            elif filtersources:
-                sp = source.SourceFilter
```
micheles: Check if it makes sense to remove weightsources

openquake/commonlib/source.py
```diff
@@ -712,223 +710,156 @@ def collect_source_model_paths(smlt):
...
-    weight = False  # when True, set the weight on each source
-
-    def __init__(self, sitecol, maxdist, monitor):
-        self.sitecol = sitecol
-        self.maxdist = maxdist
+    def __init__(self, csm, taskfunc, concurrent_tasks, maximum_distance,
+                 monitor):
+        self.tm = parallel.TaskManager(taskfunc)
+        self.csm = csm
+        self.maxweight = math.ceil(csm.weight / (concurrent_tasks or 1))
```
ptormene: Please remove it if csm.maxweight is already available

openquake/commonlib/source.py
```diff
@@ -712,223 +710,156 @@ def collect_source_model_paths(smlt):
...
+                    self.sources_by_trt.get(trt_model.id, []),
                     key=operator.attrgetter('source_id'))
+                weight += sum(src.weight for src in trt_model)
                 if not trt_model.sources:
                     logging.warn(
                         'Could not find sources close to the sites in %s '
                         'sm_lt_path=%s, maximum_distance=%s km, TRT=%s',
                         source_model.name, source_model.path,
                         self.maxdist, trt_model.trt)
+        self.csm.weight = weight
```
ptormene: Please check if we need to have two different weights, having one for splitted

ptormene:
LGTM

PR #615, opened 2016-01-07
Worked around a bug in h5py.Group.visitems in Ubuntu 12.04

micheles:
For some reason, we had a segmentation fault in the demo LogicTreeCase1ClassicalPSHA when using oq-lite on Ubuntu 12.04. The problem did not appear in other demos. The workaround is to not call DataStore.set_nbytes which in turn was calling the offending `visitems`.

The demos are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1291


daniviga:
LGTM

PR #616, opened 2016-01-11
Changed the rupture calculator to generate one seeed per each rupture

micheles:
This makes it possible to split a source in ruptures without changing the final numbers.


ptormene:
LGTM

PR #617, opened 2016-01-11
Added an end-to-end test for GMPETable

micheles:
As reported by our Canadian users, the GMPETable functionality is not actually working in the engine. There were a few bugs that are solved here. And end-to-end test has been added by reducing a real life case. The suggested way to use a GMPETable is:

1. put the .hdf5 files in the same directory where the gsim_logic_tree.xml file is
2. write in the gsim_logic_tree.xml file the names of the .hdf5 files, as follows:

```xml
    <uncertaintyModel gmpe_table="Wcrust_med_rhypo.hdf5"> GMPETable</uncertaintyModel>
```

The tests are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1301/


ptormene:
If tests are green, LGTM

PR #618, opened 2016-01-12
Used the src.num_ruptures attribute defined in hazardlib

micheles:
Companion of https://github.com/gem/oq-hazardlib/pull/412. I ran the test manually with success.


g-weatherill:
LGTM

PR #619, opened 2016-01-13
Split sources even in classical tiling

micheles:
The tiling calculator in oq-lite was not splitting the sources and thus it was distributing very poorly. Now I am splitting the heavy sources, where heavy means `src.weight > total_weight / concurrent_tasks` and the calculator is distributing well. The numbers are slightly different because of the splitting (see http://micheles.github.io/2016/01/10/to-split-or-not-to-split/)

Tests and demos pass: https://ci.openquake.org/job/zdevel_oq-risklib/1303


ptormene:
LGTM

Issue #370, opened 2015-09-14
If the time_event is wrong, we get the error too late


micheles:
This happens in the scenario_risk calculator. For instance, if the user writes"Day" instead of "day" she gets:

```python
  File "/home/michele/oq-risklib/openquake/risklib/workflows.py", line 265, in gen_out_by_rlz
    values = get_values(loss_type, assets, self.time_event)
  File "/home/michele/oq-risklib/openquake/risklib/workflows.py", line 202, in get_values
    for a in assets])
  File "/home/michele/oq-risklib/openquake/risklib/workflows.py", line 147, in value
    return self.values['fatalities_' + str(time_event)]
KeyError: 'fatalities_Day'
```

but this happens too late, in the execute phase whereas it should happen in the pre_execute phase.

PR #620, opened 2016-01-13
Added a cross validation for the `time_event` parameter

micheles:
Fixes https://github.com/gem/oq-risklib/issues/370


ptormene:
LGTM

PR #624, opened 2016-01-14
packager.sh: make CI job fails when tests are broken

daniviga:
This makes the Jenkins job to exit (with a failure) when a test is broken. It also add support for the `GEM_DEBUG` functionality.

Tests are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1308/


micheles:
LGTM

Issue #563, opened 2015-12-01
Investigate the case of inconsistent number of loss ratios


micheles:
This happens in the classical risk export test in the engine.

micheles:
After deep investigation, I discovered at least a couple of serious bugs of the classical_risk calculator, when the number of loss_ratios changes between insured and non-insured curves, or between curves of different taxonomies. Also, the need for a serious refactoring of the risk model came out stronger than before.

PR #626, opened 2016-01-19
Propedeutic refactoring of the risk models management

micheles:
This simplification is needed before we can fix the bugs in the classical_risk calculator (https://github.com/gem/oq-risklib/issues/563). This PR implements the following changes:

1. the global dictionary rmdict that was populated by side effect has been removed (it was a hack
from the beginning)
2. the event_based_bcr workflow has been removed (it was a leftover that we forgot to remove before)
3. readinput.get_risk_model accepts now two parameters and the change requires a companion PR
in the engine
4. a bug with the loading of the risk model has been fixed, now it is always loaded when the files are there (before it was not loaded if it was already available in the datastore)
5. the oqparam object has two new attributes .file_type and .risk_files which are for internal usage
6. the management of fragility functions has been made more similar to the management of risk functions 
by adding a method mean_loss_ratios_with_steps
7. we removed a special case in get_workflow

A lot more refactoring will be needed before this part of the code can be considered clear.

The tests and demos are running: https://ci.openquake.org/job/zdevel_oq-risklib/1312


ptormene:
LGTM

PR #631, opened 2016-01-20
Added a test case for vulnerability function with different curve resolutions

micheles:
I am adding the test, but skipping it. A second pull request will complete the fix. See https://github.com/gem/oq-risklib/issues/632

This is a case with two taxonomies, MC-RLSB-2 has a vulnerability function with 33 unique loss ratios
and MS-SLSB-1 with 25 unique loss ratios.


openquake/qa_tests_data/classical_risk/case_5/vulnerability_model.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0"?>
+<nrml xmlns="http://openquake.org/xmlns/nrml/0.4" xmlns:gml="http://www.opengis.net/gml">
```
raoanirudh: Please update to the current format to be consistent with the other test cases.

raoanirudh:
Apart from the format of the vulnerability model, LGTM.

openquake/qa_tests_data/classical_risk/case_5/vulnerability_model.xml
micheles: Will do

Issue #632, opened 2016-01-20
Make sure the classical_risk calculator is able to manage vulnerability functions with a different number of loss ratios


micheles:
This is the last feature required before we can drop the old calculator.

PR #633, opened 2016-01-21
Fixed the case of risk models with different curve resolutions

micheles:
Fixes https://github.com/gem/oq-risklib/issues/632. I have checked the numbers with the ones produced by the engine calculator in release 1.7, and they match. Anirudh is double-checking them with an independent implementation. This PR must enter together with https://github.com/gem/oq-engine/pull/1921.

All of Anirudh's test are running except case_master that will be fixed later, when the insured losses will be computed again.


raoanirudh:
LGTM

openquake/calculators/base.py
```diff
@@ -54,6 +54,19 @@ class AssetSiteAssociationError(Exception):
 rlz_dt = numpy.dtype([('uid', (bytes, 200)), ('weight', float)])
 
 
+def set_array(longarray, shortarray):
+    """
+    :param longarray: an array of length L >= l
+    :param shortarray: an array of length l
```
ptormene: Please specify in the docstring that those should be numpy arrays. For lists, for instance, it breaks.

ptormene: And for numpy arrays, for some reason I get numbers instead of nan

ptormene:
Please double-check about the `set_array` function. Apart from that, LGTM

openquake/calculators/base.py
micheles: This happens if you use arrays of integers instead of array of floats

ptormene:
LGTM

Issue #637, opened 2016-01-22
Unify the calculators classical and classical_tiling


micheles:
Now it is possible. Perhaps the classical calculator should automatically perform tiling when there are more than sites_per_tile sites.

PR #643, opened 2016-01-26
Unified classical and classical_tiling calculator

micheles:
Closes https://github.com/gem/oq-risklib/issues/637.

NB: there was a bug in the previous implementation and the file case_2/expected/hazard_map-mean.csv was incorrect. The one committed here is correct, i.e. identical to the result without tiling.


ptormene:
LGTM

PR #645, opened 2016-01-27
Remove `specific_assets` from event based risk demo

raoanirudh:
This parameter was removed in https://github.com/gem/oq-risklib/pull/513


micheles:
LGTM

PR #646, opened 2016-01-27
rpm/spec: add the new python-prctl dep (requires GEM copr repo)

daniviga:
Requires: https://copr.fedorainfracloud.org/coprs/gem/openquake/package/python-prctl/


micheles:
LGTM

PR #647, opened 2016-01-28
Store the risk models in matrix form

micheles:
This is preliminary to removing the risk functions as pickled objects: https://github.com/gem/oq-risklib/issues/648.
To check: run case_master, open the HDF5 file and look inside the `composite_risk_model` group.
Then check:

1. the case of VulnerabilityFunction with PMF
2. the case of continuous fragility functions
3. the case of discrete fragility functions

Consequence models and retrofitted models are not done yet.


raoanirudh:
LGTM.

Some suggestions for future refinements:

1. In the case of continuous fragility functions, it might be better to store the mean and stddev for each damage state instead of the discretised exceedance probabilities.
2. In the case of benefit-cost calculations, it would be nice to have the retrofitted vulnerability functions also visible in the HDF5
3. If the plan is to remove the erstwhile risk_model, it would be useful to add all of the other parameters such as `noDamageLimit`, `minIML`, and `maxIML` also as attributes

PR #651, opened 2016-01-28
Removed duplication in the usage of psutil

micheles:
Let's use the wrapper functions in baselib to work with all versions of psutil.


daniviga:
LGTM if all tests are green

Issue #487, opened 2015-10-29
Can we replace fatalities with occupation?


micheles:
That would solve the long standing confusion between loss type and cost type (see `riskmodels.cost_type_to_loss_type`) To be asked to the risk team.

raoanirudh:
Yes, though with `occupants` and not ~~`occupation`~~.

PR #654, opened 2016-01-29
Renamed fatalities -> occupants

micheles:
This fixes a hidden bug in engine-1.7 so that the loss maps for fatalities are not computed due to the
confusion `fatalities <-> occupants` (yet another one).
Closes https://github.com/gem/oq-risklib/issues/487


raoanirudh:
LGTM

Issue #648, opened 2016-01-28
Remove the risk models as pickled objects


micheles:
This is hard, since there are 5 different situations (fragility discrete, fragility continuous, vulnerability, vulnerability with PMF, vulnerability retrofitted) that must be managed differently.

PR #655, opened 2016-01-29
Removed the pickled riskmodel from the datastore

micheles:
Closes https://github.com/gem/oq-risklib/issues/648. Notice that

1. In the case of continuous fragility functions, we are storing just mean and stddev for each damage state and not the probabilities
2. In the case of benefit-cost calculations, the retrofitted vulnerability functions are also visible in the HDF5
3. other parameters such as noDamageLimit, minIML, and maxIML are also visible as attributes

i.e. we addressed all of the points raised in https://github.com/gem/oq-risklib/pull/647. The demos are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1329/


raoanirudh:
LGTM!

PR #656, opened 2016-01-29
Fixed expected files

micheles:
The master broke https://ci.openquake.org/job/master_oq-risklib/1882 since I forgot to change the expected files after the renaming fatatities -> occupants.


micheles:
LGTM

PR #657, opened 2016-01-29
Adds support for Arbitrary MFD and Youngs & Coppersmith MFD class

g-weatherill:
Two MFDs exist in the hazardlib are not currently exposed to the users in the xml: the Arbitrary Magnitude Frequency distribution and the Youngs & Coppersmith (1985) magnitude frequency distribution. This PR adds the support to read and write these models from xml.


openquake/commonlib/tests/source_test.py
```diff
@@ -34,13 +34,15 @@
 from openquake.commonlib.source import parse_source_model, DuplicatedID
 from openquake.commonlib.nrml import nodefactory
 from openquake.commonlib.node import read_nodes
-from openquake.baselib.general import assert_close
+from openquake.baselib.general import assert_close, assert_close_seq
```
micheles: `assert_close_seq` is really an internal function. You should just use `assert_close`.

micheles:

mmpagani:
LGTM

PR #658, opened 2016-02-01
Reduced classical_risk_test/case_3

micheles:
The test was inconveniently slow, so I have reduced the source model from 398 sources to 15 sources.
The runtime went down from 191 s to 11 s (on Jenkins).


ptormene:
LGTM

PR #665, opened 2016-02-01
Check the loss_maps in classical_risk/test_case_4

micheles:
The loss maps are easier to check than the loss curves, should the test break.


ptormene:
LGTM

Issue #639, opened 2016-01-23
Large SESCollections cannot be unpickled


micheles:
We saw this problem a few times, in cases with more than half million ruptures:

```python
  File "/usr/local/openquake/oq-risklib/openquake/calculators/event_based.py", line 582, in pre_execute
    for col_id, sescol in enumerate(self.datastore['sescollection']):
  File "/usr/local/openquake/oq-risklib/openquake/commonlib/datastore.py", line 263, in __getitem__
    val = pickle.loads(val.value)
EOFError
```

Probably there is a limit on how long a string can be in HDF5, so the pickle byte string is truncated and cannot be read back. One lighweight solution is to split the sescollection (i.e. we could have a sescollection per source), but the real solution is to avoid storing the ruptures as pickled objects. Instead, we should store the Context objects as numpy arrays.

PR #666, opened 2016-02-01
Stored the SESCollection as a set of pickled objects instead of a single one

micheles:
This should help to solve https://github.com/gem/oq-risklib/issues/639 temporarily. It is important for the California project, so that Graeme can generate one collection per branch. The sescollections (each sescollection is a dictionary {tag: SESRupture} are stored by collection key (`colkey`) which is a string of the form `trtmod=<trt_id>-<idx>` where `trt_id` is the tectonic model index and `idx` is always zero
in the case of full enumeration. If sampling is enabled then `idx` is the sampling ordinal of the model.
For instance, in the test case_17, which has 2 tectonic region models and `number_of_logic_tree_samples = 5`, the SESCollection has the following 5 components:

sescollection/trtmod=0-0
sescollection/trtmod=1-0
sescollection/trtmod=1-2
sescollection/trtmod=1-3
sescollection/trtmod=1-4

### DIFF
--- a/openquake/calculators/event_based.py
+++ b/openquake/calculators/event_based.py
@@ -383,7 +383,6 @@ class EventBasedRuptureCalculator(ClassicalCalculator):
     """
     core_task = compute_ruptures
     tags = datastore.persistent_attribute('tags')
-    sescollection = datastore.persistent_attribute('sescollection')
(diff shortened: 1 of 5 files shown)


g-weatherill:
LGTM

PR #670, opened 2016-02-02
Fixed the initial value of the accumulator in EventBasedRuptureCalculator

micheles:
Before this fix the result of the EventBasedRuptureCalculator was a dictionary with sometimes a scalar key (the `trt_model_id`) and sometimes a pair `(trt_model_id, gsim)`. This was very annoying and required some escamotage like the `isinstance(trt_id, tuple): continue` below. Now the key is always the integer `trt_model_id`.


ptormene:
LGTM

Issue #669, opened 2016-02-02
Serialize the CompositionInfo


micheles:
This requires to store three attributes: seed, num_samples, gsim_lt_xml

PR #672, opened 2016-02-03
Serialized the CompositionInfo object to and from HDF5

micheles:
Closes https://github.com/gem/oq-risklib/issues/669


ptormene:
LGTM

PR #673, opened 2016-02-04
Added a SourceModelParser with caching

micheles:
This is an optimization useful in cases of large source model logic trees, when a specific source model file has to be considered several times. For instance in our demo [LogicTreeCase2ClassicalPSHA](https://github.com/gem/oq-risklib/blob/engine-1.7/demos/hazard/LogicTreeCase2ClassicalPSHA/report.rst) the file `source_model.xml` is considered 81 times.
In the current master we parse the same file 81 times and we re-instantiate and re-validate the
sources 81 times; with this pull request the file will be parsed only once and the sources validated only once, then we will make 81 copies of the sources and we will apply the uncertainties to the copies.

This makes a difference especially for the command ``oq-lite info --report``. In our demo the time to read the composite source model goes down from 12s to 6s. In more complex cases the speedup can be even bigger.

PS: while working at this feature I discovered a few subtle bugs in the Node class which have been fixed.

The demos are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1336/

### DIFF
--- a/openquake/commonlib/tests/sourcewriter_test.py
+++ b/openquake/commonlib/tests/sourcewriter_test.py
@@ -39,12 +39,14 @@ class SourceWriterTestCase(unittest.TestCase):
         self.check_round_trip(ALT_MFDS)
 
 
-# deep-copied sources must be serializable
 class DeepcopyTestCase(unittest.TestCase):
-    def test(self):
+    def test_is_writeable(self):
         parser = SourceModelParser(SourceConverter(50., 1., 10, 0.1, 10.))
         [sf, cf] = map(copy.deepcopy, parser.parse_sources(ALT_MFDS))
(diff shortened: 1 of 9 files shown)


openquake/commonlib/tests/sourcewriter_test.py
```diff
@@ -48,3 +37,14 @@ def test_nonparam(self):
 
     def test_alt_mfds(self):
         self.check_round_trip(ALT_MFDS)
+
+
+# deep-copied sources must be serializable
+class DeepcopyTestCase(unittest.TestCase):
```
ptormene: Please write more explicitly where things went wrong before adding the `__deepcopy__`

ptormene:
LGTM

Issue #551, opened 2015-11-25
For scenario_damage, it is okay to have an exposure without costs


micheles:
Instead now we get an error like this (discovered by Catalina):

```python
ValueError: node asset: Invalid Exposure. Missing cost set(['structural']) for asset ant_1_M1, line 10 of /home/cyepes/Antioquia-scenario-Nov25/exposure_model_Antioquia_v3.xml
```

PR #677, opened 2016-02-04
Removed a check on the costs too strict for damage calculations

micheles:
Closes https://github.com/gem/oq-risklib/issues/551


raoanirudh:
LGTM

PR #679, opened 2016-02-05
Enhanced the reports

micheles:
Now the effective number of ruptures is reported (the concept will be documented in the future) as well as the processing times for the slowest sources. Moreover the estimated logic tree is closer to the reality since the source filtering is taken into account.

The demos are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1354/


openquake/commonlib/source.py
```diff
@@ -1039,17 +1067,25 @@ def store_source_info(self, dstore, task_name):
 
 
 @parallel.litetask
-def dummy_task(sources, sitecol, siteidx, rlzs_assoc, monitor):
-    return {}
+def count_eff_ruptures(sources, sitecol, siteidx, rlzs_assoc, monitor):
+    """
+    Count the number of ruptures contained in the given sources and return
+    a dictionary trt_model_id -> num_ruptures. All sources belongs to the
```
ptormene: Typo `belongs`->`belong`

openquake/commonlib/source.py
```diff
@@ -536,6 +554,17 @@ def __repr__(self):
...
 
 
+def cjoin(strings):
+    """
+    :param strings: an iterable over strings
+    :returns: a comma separated string of lenght <= LENGHT
+    """
+    s = ','.join(strings)
+    if len(s) > LENGTH:
+        raise ValueError('The string %r is over %d characters' % (s, LENGTH))
```
ptormene: Please add a comment to explain why a string longer than `LENGTH` would cause problems, and in what kind of situation it might happen.

openquake/commonlib/reportwriter.py
```diff
@@ -35,13 +35,15 @@ class ReportWriter(object):
...
+        short_source_info='Slowest sources',
+        performance='Slowest operations',
     )
 
     def __init__(self, dstore):
         self.dstore = dstore
         self.oq = oq = OqParam.from_(dstore.attrs)
         self.text = oq.description + '\n' + '=' * len(oq.description)
-        sitecol_size = humansize(len(dstore.hdf5['sitecol'].value))
+        sitecol_size = humansize(len(parallel.Pickled(dstore['sitecol'])))
```
ptormene: Please add a comment, specifying that pickling is currently the right way to estimate the data transfer, but we could change it if we transfer the sitecol as an array

openquake/calculators/views.py
```diff
@@ -172,6 +175,11 @@ def view_ruptures_per_trt(token, dstore):
     return rst_table(tbl, header=header) + summary
 
 
+@view.add('short_source_info')
+def view_short_source_info(token, dstore):
+    return rst_table(dstore['source_info'][:20])
```
ptormene: Please, instead of 20, use an additional parameter, with 20 as default

openquake/calculators/views.py
```diff
@@ -150,20 +151,22 @@ def view_ruptures_per_trt(token, dstore):
                 lambda rows: sum(r['weight'] for r in rows))
     n = groupby(source_info, operator.itemgetter('trt_model_id'),
                 lambda rows: sum(1 for r in rows))
-    for sm in csm_info.source_models:
+    for i, sm in enumerate(csm_info.source_models):
+        erdict = dict(ast.literal_eval(csm_info.eff_ruptures[i]))
```
ptormene: Please add a comment to explain this line

openquake/commonlib/source.py
micheles: Fixed

openquake/commonlib/source.py
micheles: I have reduced the error to a warning, since even if happens (which is extremely unlikely) it would break the .rst report, not the computation.

openquake/commonlib/reportwriter.py
micheles: Done

openquake/calculators/views.py
micheles: Done

openquake/calculators/views.py
micheles: Done

openquake/commonlib/source.py
```diff
@@ -536,6 +554,24 @@ def __repr__(self):
...
+# overflow would break the report but not the computation
+def cjoin(strings, maxlength=LENGTH):
+    """
+    Join strings by checking that the result is below maxlength characters
+
+    :param strings: an iterable over strings
+    :returns: a comma separated string of lenght <= maxlength
+    """
+    s = ','.join(strings)
+    if len(s) > LENGTH:
```
ptormene: You forgot to change `LENGTH -> maxlength` also here

openquake/commonlib/source.py
```diff
@@ -695,8 +730,8 @@ def get_rlzs_assoc(self, count_ruptures=lambda tm: tm.num_ruptures):
                 nr = count_ruptures(tm)
                 if nr:
                     trts.append(tm.trt)
-                    rups.append(str(nr))
-            self.eff_ruptures[i] = ','.join(rups)
+                    rups.append('%d: %d' % (tm.id, nr))
+            self.eff_ruptures[i] = '{%s}' % cjoin(rups, LENGTH - 2)
```
ptormene: What does 2 mean here?

ptormene:
LGTM

PR #685, opened 2016-02-08
Added the ability to override the job.ini parameters from the command-line

micheles:
This is very usuful, for instance to check the seed dependency. Try for instance

```
$ oq-lite run job.ini --param random_seed=142
```

or

```
$ oq-lite run job.ini --param investigation_time=1000 ses_per_logic_tree_path=10
```


raoanirudh:
LGTM!

PR #676, opened 2016-02-04
Introduced a cutoff for the IML=0 situation caused by some fragility functions

micheles:
Closes https://github.com/gem/oq-risklib/issues/584. The cases `noDamageLimit=0` and `minIML=0` are errors, but they are presents in lots of our fragility functions, so we will use a "hide under the carpet" approach to avoid the annoying divideByZero warnings raised by hazardlib.

PR #687, opened 2016-02-09
Check if there are PoE=1 in classical_damage calculations

micheles:
Closes https://github.com/gem/oq-risklib/pull/676


micheles:
Superseded by https://github.com/gem/oq-risklib/pull/692

PR #689, opened 2016-02-10
Use the realization ordinal as key in the datastore for asset_loss_table and agg_loss_table

micheles:
Also, I fixed a long standing bug with the `ses_ratio` when sampling is enabled and remove a not-so-interesting warning. Even the demos are running fine: https://ci.openquake.org/job/zdevel_oq-risklib/1361


ptormene:
LGTM

PR #692, opened 2016-02-10
Raised an error early if there is a PoE=1 for classical damage

micheles:
Closes https://github.com/gem/oq-risklib/pull/676


raoanirudh:
LGTM

PR #694, opened 2016-02-12
Copyright 2016 and sources headers grand refactoring

nastasi-oq:
tests are running: https://ci.openquake.org/job/zdevel_oq-risklib/1363/


daniviga:
Do we want to add an item in the changelog as we have done on hazardlib?

LGTM

PR #695, opened 2016-02-17
Replaced some DataStore with datastore.read

micheles:
This should solve the problem in https://ci.openquake.org/job/master_oq-engine/2254/console
The demos are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1368


daniviga:
LGTM

PR #696, opened 2016-02-18
Made the header of CSV files customizable

micheles:
It is now possible to use a shorter alias instead of the full column description, as requested by the risk team.


raoanirudh:
LGTM

PR #698, opened 2016-02-19
Export specific GMFs

micheles:
For scenario calculations the risk team wants to be able to export a specific GMF as a CSV file with header (lon, lat, value-first-IMT, ... value-last-IMT). I have extended the export commands as follows:

```bash
$ oq-lite export gmfs:<n>
```

will export the GMF corresponding to the n-th realization (n is in the range 0 .. N - 1 where N is the total number of realizations). It is also possible to export more than one realization; for instance

```bash
$ oq-lite export gmfs:0,1,2
```

will export the first three realizations. The name of the files is `gmf-<rupture-tag>.csv`.
If the scenario calculation has a non-trivial logic tree (i.e. there is more than one GSIM) multiple files are exported, with names of the form `gmf-<GSIM>-<rupture-tag>.csv`.

NB: several numbers have changed slightly, since I have reduced the precision of the GMFs from 64 bit to 32 bit.

Tests and demos are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1373


openquake/calculators/scenario_damage.py
```diff
@@ -146,7 +146,10 @@ def pre_execute(self):
         base.RiskCalculator.pre_execute(self)
         self.monitor.consequence_models = riskmodels.get_risk_models(
             self.oqparam, 'consequence')
-        _, gmfs = base.get_gmfs(self)
+        if 'gmfs' in self.oqparam.inputs:
+            self.rlzs_assoc = logictree.trivial_rlzs_assoc()
+        self.tags, gmfs = base.get_gmfs(self.datastore)
+        # self.save_params()  # save number_of_ground_motion_fields and sites
```
ptormene: Should this line be removed?

ptormene:
LGTM

openquake/calculators/scenario_damage.py
micheles: Yes, I am removing it

Issue #463, opened 2015-10-23
Change the risk CSV exports by following the document of Catalina and Anirudh


micheles:
https://docs.google.com/spreadsheets/d/1XQzexMmlD4I_k5dgo-bnkalFD2nUVeNFsQhqVvH9rkA/edit?ts=560ba112#gid=369943397

raoanirudh:
### Template for scenario losses per asset

---

File path and name:

if single gsim: **_`export_dir`/losses-`loss_type`.csv_**
else: **_`export_dir`/`gsim_name`/losses-`loss_type`.csv_**

---

| asset_ref | taxonomy | longitude    | latitude    | mean_loss | stddev_loss | mean_insured_loss | stddev_insured_loss |
|-----------|----------|--------------|-------------|-----------|-------------|-------------------|---------------------|
| a1        | concrete | -1.22000E+02 | 3.81130E+01 | 3.210E+01 | 2.100E+01   | 1.926E+01         | 1.260E+01           |
| a2        | concrete | -1.22114E+02 | 3.81130E+01 | 1.988E+01 | 1.212E+01   | 1.193E+01         | 7.274E+00           |
| a3        | steel    | -1.22570E+02 | 3.81130E+01 | 1.520E+01 | 1.934E+01   | 9.120E+00         | 1.160E+01           |
| a4        | masonry  | -1.22000E+02 | 3.80000E+01 | 3.537E+01 | 2.314E+01   | 2.122E+01         | 1.389E+01           |
| a5        | adobe    | -1.22000E+02 | 3.79100E+01 | 2.191E+01 | 1.336E+01   | 1.315E+01         | 8.016E+00           |
| a6        | concrete | -1.22000E+02 | 3.82250E+01 | 1.675E+01 | 2.131E+01   | 1.005E+01         | 1.279E+01           |
| a7        | wood     | -1.21886E+02 | 3.81130E+01 | 3.898E+01 | 2.550E+01   | 2.339E+01         | 1.530E+01           |
| ⋮         | ⋮        | ⋮            | ⋮           | ⋮         | ⋮           | ⋮                 | ⋮                   |

---

**Comment**: If insured losses are not computed, the last two columns in the table should be skipped.

raoanirudh:
### Template for scenario loss map

---

File path and name:

if single gsim: **_`export_dir`/loss-map-`loss_type`.csv_**
else: **_`export_dir`/`gsim_name`/loss-map-`loss_type`.csv_**

---

[...]

PR #699, opened 2016-02-22
Changed the CSV exporter for the loss maps

micheles:
Now I am re-using the same CSV exporter used by the classical_risk and event_based calculators. There is a simplification, and also the export format is closer to what the risk team wants (https://github.com/gem/oq-risklib/issues/463). I am also reducing the default precision to 6 digits after the dot, since now the GMFs are limited to 32 bits.

The demos are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1381


raoanirudh:
LGTM

Issue #405, opened 2015-09-30
Add csv exporters for the scenario_damage outputs


micheles:
Now I have specs from Catalina and Anirudh, https://docs.google.com/spreadsheets/d/1XQzexMmlD4I_k5dgo-bnkalFD2nUVeNFsQhqVvH9rkA/edit?ts=560ba112#gid=535998355

raoanirudh:
### Template for scenario ground motion fields

|   Longitude  |   Latitude  |  PGA_#0001  |  PGA_#0002  |  PGA_#0003  | … | SA(0.1)_#0001 | SA(0.1)_#0002 | SA(0.1)_#0003 | … |
|:------------:|:-----------:|:----------:|:----------:|:----------:|:-:|:------------:|:------------:|:------------:|:-:|
| -1.22570E+02 | 3.81130E+01 | 6.9481E-01 | 3.3424E-01 | 1.2495E-01 | … |  2.3295E-01  |  6.4395E-01  |  5.0039E-01  | … |
| -1.22114E+02 | 3.81130E+01 | 1.0090E-01 | 2.1241E-01 | 2.1322E-02 | … |  5.9193E-01  |  3.6746E-01  |  1.9361E-01  | … |
| -1.22000E+02 | 3.81130E+01 | 2.7647E-01 | 9.2468E-01 | 5.3349E-01 | … |  2.0677E-01  |  9.2979E-01  |  2.1163E-01  | … |
| -1.22000E+02 | 3.80000E+01 | 6.7521E-01 | 8.1728E-01 | 3.5368E-01 | … |  9.1544E-01  |  5.6278E-01  |  8.3956E-01  | … |
| -1.22000E+02 | 3.79100E+01 | 5.8284E-01 | 8.7594E-01 | 6.9020E-01 | … |  1.8499E-01  |  8.0060E-01  |  5.0044E-01  | … |
| -1.22000E+02 | 3.82250E+01 | 4.1964E-01 | 4.7037E-01 | 6.1432E-01 | … |  2.0335E-01  |  7.4245E-01  |  6.1753E-01  | … |
| -1.21886E+02 | 3.81130E+01 | 8.9685E-01 | 1.6730E-01 | 1.6314E-01 | … |  3.0965E-01  |  2.1864E-01  |  6.3308E-01  | … |
[...]

PR #702, opened 2016-02-23
Export GMFs by IMT

micheles:
As requested by the risk team (https://github.com/gem/oq-risklib/issues/405). The demos are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1386

PS: while at it, I am also introducing a progressive serial number for each rupture, starting from 1 and saved in the array `src.serial`. That makes the rupture tag a lot more descriptive and for instance `rup=10142-03` now means third sample of the 10,142th rupture of the source model.


openquake/commonlib/export/hazard.py
```diff
@@ -604,23 +605,88 @@ def export_gmf_spec(ekey, dstore, spec):
...
+        for (trt, gsim), gmfs_ in gmfs_by_trt_gsim.items():
+            for imt in gmfs_.dtype.names:
+                gmfs = numpy.zeros(len(gmfs_), dt)
+                for i, gmf in enumerate(gmfs_):
+                    gmfs[i] = tuple(gmfs_[imt][i])
+                dest = dstore.export_path('gmf-%s-%s.csv' % (gsim, imt))
+                data = util.compose_arrays(sitemesh, gmfs)
+                writer.save(data, dest)
+    else:  # event based
+        return []
```
ptormene: Please add a warning in this case

ptormene:
LGTM

PR #703, opened 2016-02-23
Made possible to generate the report even for descriptions with non-ASCII characters

micheles:
I discovered the issue in Anirudh's California calculation: while the calculation runs, and even

```bash
$ oq-lite show fullreport
```

works (because the default encoding for the terminal is UTF-8),

```bash
$ oq-lite show fullreport > report.rst
```

fails with an error

```python
UnicodeEncodeError: 'ascii' codec can't encode character u'\u2014' in position 23: ordinal not in range(128)
```

The solution is to encode the description as UTF-8 explicitly. I am adding a non-ASCII dash character to the AreaSource demo to test the report generation.

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,4 +1,6 @@
   [Michele Simionato]
+  * Made possible to generate the report even for descriptions with
+    non-ASCII characters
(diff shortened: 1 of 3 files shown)


daniviga:
LGTM

PR #706, opened 2016-02-25
Saving space when there are no insured losses

micheles:
For instance in the case with 1 loss type the space for each asset loss record is reduced from

(rup_idx, ass_idx, loss, loss_ins) -> (rup_id, ass_id, loss)

i.e. from

4 + 4 + 4 + 4 = 16 -> 4 + 4 + 4 = 12 bytes 

i.e. a 25% saving. In practice the improvement is minor but at least the form of the exported files is nicer, i.e. we avoided exporting columns full of NaNs in the case of no insured losses.

PS: the 25% save have been checked in a computation from South East Asia. The asset loss table in the HDF5 file was reduced from 4.41 GB to 3.31 GB.


raoanirudh:
LGTM

PR #707, opened 2016-02-26
Discriminating does not exist versus cannot access

micheles:
Daniele was right and `os.path.exists` says that a file does not exist when it does not have permission to read it, even it exists. Fortunately `open` works correctly, so I am using that instead.


daniviga:
LGTM

PR #708, opened 2016-02-26
Fixed the case with statistics and no conditional_loss_poes

micheles:
I discovered this error while running the Indonesia computation:

```python
     File "/usr/local/openquake/oq-risklib/openquake/calculators/event_based_risk.py", line 553, in compute_store_stats
       loss_maps = numpy.zeros((self.N, self.Q1), self.loss_maps_dt)
   TypeError: Empty data-type
```

The solution was to add a couple of `if conditional_loss_poes`. I also made test_case_3 sensitive to this issue by simply producing some loss curves.


daniviga:
LGTM

PR #710, opened 2016-02-26
Simplified method gen_out_by_rlz

micheles:
This is a small refactoring. The management of missing values it transferred to the ScenarioRisk calculator, i.e. the only place where it is needed. Then we can unify the two different versions of the method gen_out_by_rlz and save a few lines of code.

The demos are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1392


ptormene:
If demos are green, LGTM

PR #713, opened 2016-02-28
Saved data transfer information in event based risk

micheles:
This is essential to find the data transfer bottlenecks.
PS: while at that, I am also changing the display of the datastore contents, to show the biggest ones at the end.


ptormene:
LGTM

Issue #715, opened 2016-02-29
After running the demo LogicTreeCase1ClassicalPSHA it is impossible to export the UHS


micheles:
There is a confusion in the case of a single PoE. Probably we should change the composite array to avoid subarrays. This is also useful for the future, when UHS will be defined for IMTs != spectral acceleration.

NB: `$ oq-engine --run job.ini --exports xml` does not export the UHS, this is why the problem was
not discovered earlier.

PR #718, opened 2016-03-01
Fixed the export of UHS curves

micheles:
It was failing in the case of a single PoE. I have also change the CSV format to be a real CSV (i.e. no space-separated composite fields anymore). Closes https://github.com/gem/oq-risklib/issues/715


ptormene:
If tests are green, LGTM

PR #725, opened 2016-03-02
Trap errors when filtering sources

micheles:
We should print the `source_id` of the source causing issues. This was discovered while investigating https://github.com/gem/oq-engine/issues/1965.


ptormene:
LGTM

ptormene:
LGTM

PR #726, opened 2016-03-03
Doing the aggregation by asset in numpy, not in Python

micheles:
This is a lot faster. For instance on gemsun02 the runtime of our demo went down from 395 s to 208 s and  the memory allocation from 30 MB to 22 MB. In the case of Turkey with 100 years I saw a change from  81,373 s -> 47,264 s.

The demos are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1400/

### DIFF
--- a/openquake/calculators/event_based_risk.py
+++ b/openquake/calculators/event_based_risk.py
@@ -24,6 +24,7 @@ import collections
 
 import numpy
 
+from openquake.baselib.python3compat import zip
 from openquake.baselib.general import AccumDict, humansize
 from openquake.calculators import base
 from openquake.commonlib import readinput, parallel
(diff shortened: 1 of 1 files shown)


openquake/calculators/event_based_risk.py
```diff
@@ -179,6 +173,12 @@ def zeroN():
                         monitor.ela_dt))
 
             # agg losses
+            glosses = out.event_loss_per_asset.sum(axis=1)
+            ilosses = out.insured_loss_per_asset.sum(axis=1)
+            agglosses = {
+                rupid: numpy.array([gloss, iloss], F32)
+                for rupid, gloss, iloss in zip(out.rupids, glosses, ilosses)
```
ptormene: The same thing is called in a number of different ways (`all_losses, event_loss_per_asset, groundloss, glosses`). Perhaps we should attempt to unify them.

ptormene:
Apart from a minor note about naming conventions, LGTM

openquake/calculators/event_based_risk.py
micheles: You have a point here, but since I have another branch were all the names have changed again, I will wait for the other pull request.

Issue #729, opened 2016-03-05
Risk by site


micheles:
Typically risk computations have more than one asset per hazard site. For instance, consider a case with 1 million of assets and 10,000 sites, with 100 assets per each site (think of California). If we split the computation by site, the maximum amount of memory needed is the memory needed to store the output of 100 sites, i.e. 10,000 times less than the memory that we are using now! Actually things are more complex: on one side, we still need to keep the GMFs in memory for all sites (but this surprisingly does not amount to much), on the other side the calcution is already split by taxonomy, so the memory occupation can be even lower: the limit is the maximum number of assets per site per taxonomy, which can be quite low. In practice doing the risk by site will solve our memory problems in the event based calculator: on top of that, it will make it possible to further optimize the calculation. Right now we are repeating computations that should not be repeated: for instance we are computing the probabilities from the vulnerability functions N times instead of once, when N is the number of assets of the same taxonomy on a given site.

PR #730, opened 2016-03-05
Removed risklib.qa_tests

micheles:
I am removing a lot of redundant annoying tests. Such tests are sensitive to the details of the current implementation and are actually harmful, since they impede refactoring. On the other hand everything is covered much better with end-to-end tests, so they are now useless (they made sense three years ago, when there was no coverage with end-to-end tests). They are getting in the way for https://github.com/gem/oq-risklib/issues/729.


daniviga:
LGTM

PR #736, opened 2016-03-09
Export all the SES collections in a single file

micheles:
Instead of one file per source model logic tree path and tectonic region model. This was requested by Graeme.


g-weatherill:
LGTM

PR #739, opened 2016-03-14
setup.py: remove upper limit in psutil

daniviga:
Upper limit is blocking port to Ubuntu 16.04. The first tests I did with Ubuntu 16.04 look fine, if we'll found issues with psutil > 3.0 we will need to fix the code anyway.


nastasi-oq:
LGTM

PR #741, opened 2016-03-16
Simplified the event based risk test case_2bis

micheles:
To make it easier to debug when it breaks. Now there are only two assets and ten ruptures.


ptormene:
LGTM

PR #742, opened 2016-03-16
Fixed psutil wrong call in writers_test.py

micheles:
The direct call to `proc.get_memory_info()` was a slip, since in the same file in other places the right call to `memory_info` is done.


ptormene:
LGTM

Issue #747, opened 2016-03-20
Use the hazardlib fast way in case of ruptures occurring multiple times


micheles:
The GmfComputer in hazardlib has a fast method (https://github.com/gem/oq-hazardlib/blob/engine-1.9/openquake/hazardlib/calc/gmf.py#L95) to compute multiple GMFs at the same time with a single seed, however the engine is using the slow method https://github.com/gem/oq-hazardlib/blob/engine-1.9/openquake/hazardlib/calc/gmf.py#L168 that requires a seed for each GMF. The reason is that the slow method is parallelizable whereas the fast method is not. However, reality tells us that the fast method is faster than the parallelizable method even if you have several cores. For instance, on my workstation with 16 tasks,   the "compute poes" time for the ScenarioRisk demo with 100,000 events goes down from 560 cumulative seconds (equivalent to 560/16 = 35 s) to 12 seconds, so the non-parallelizable method is still faster. This is even more so in case of correlation: for the demo the time goes down from 203,282s to 112s (1815x, even divided by 16 is 113x). The memory consumption clearly increases (from 68 MB to 546 MB, 8x) but in a manageable way.

Because of the California calculation with correlated GMFs, it has become imperative to use the fast method in event based calculations, where there is no parallelization issue, since we parallelize by ruptures, not by events. In theory we could use the fast method in event based calculations and the parallelizable method in scenario calculations, but I prefer to use the same approach everywhere, for sake of simplicity. Even if the parallelizable method could save few minutes in the hazard part of a scenario calculation, it is nothing compared to the hours spent in the risk part, so it is not worth the optimization effort.

NB: using the fast method is akin to a change of seeds and therefore most expected files in the test
will have to be changed.

NB: just to give an idea of the speedups involved, for Miriam's island on Wilson the calculation of the GMFs ("compute poes") goes down from 3014s to 108s (28x) in the case with no correlation.

PR #748, opened 2016-03-21
Use the fast method calcgmfs in case of ruptures occurring multiple times

micheles:
Closes https://github.com/gem/oq-risklib/issues/747


raoanirudh:
LGTM

PR #751, opened 2016-03-22
Fix the management of celery

micheles:
Closes https://github.com/gem/oq-engine/issues/1987. Now celery is managed directly in commonlib.parallel, but it is still optional and imported only if the environment variable OQ_DISTRIBUTION is set to the string "celery". The demos are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1428


openquake/commonlib/parallel.py
```diff
@@ -35,20 +35,34 @@
...
 executor = ProcessPoolExecutor()
 # the num_tasks_hint is chosen to be 2 times bigger than the name of
 # cores; it is a heuristic number to get a good distribution;
 # it has no more significance than that
 executor.num_tasks_hint = executor._max_workers * 2
 
+OQ_DISTRIBUTE = os.environ.get('OQ_DISTRIBUTE', 'futures').lower()
+
+
+if OQ_DISTRIBUTE == 'celery':
```
daniviga: To be able to use `celery` the celeryconfig must be imported before any import from celery:

```python
try:
    import celeryconfig
except ImportError:
     sys.path.append('/usr/share/openquake/engine')
```

micheles: Done

daniviga:
LGTM is tests are now green

PR #754, opened 2016-04-01
Adds index method to Node object

g-weatherill:


micheles:
I would like to keep the namespace of the `Node` class limited. The problem is that if you had an `<index>` tag in the XML, `node.index` would call the index method and not the index node, as one could expect. There is an easy solution: make `index(node, tag)` a function, not a method.

PR #755, opened 2016-04-04
Moved the asset_refs into a different array

micheles:
Most of the asset collection array is used to store the names of the assets (asset_refs) as strings with 100 bytes each, even if the name is shorter. With this PR I move the asset_refs to a different array and keep in the asset collection only the indexes to the names. This reduces the memory occupation from 100 bytes per asset to 4 bytes per asset. The change has an impact on the data transfer in computations with a large number of assets. For instance for California with 240,000 assets I get:

master branch: `assetcol size: 30.2 MB`
asset_refs branch: `assetcol size: 8.2 MB`

PS: while at it, I removed some dead code related to the `specific_assets` functionality that has been dropped and I did the renaming `idx -> ordinal`  so that `idx` is now free to store the index to the asset_ref.

The demos are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1440/


ptormene:
LGTM

Issue #753, opened 2016-03-31
Permit the use of the "." and "|" characters in the source IDs


g-weatherill:
In implementing the US National Seismic Hazard map for the gridded seismicity source there is nothing to uniquely identify a point besides its location. In this case it can be useful to have the ID as the location string (e.g. `10.0N_31.0E`) or as a join of several variables based on conditions. The "." and "|" characters would allow some flexibility in the representation

micheles:
Do we really need the '|' character? In the past it gave us problems since it is not a valid characters for Windows filenames. That means that it would be impossible to save data about the source in a file with the same name as the source on Windows in presence of a '|'.

g-weatherill:
Probably it is not critical to have the "|". What about an alternative separator (e.g. "~")?

micheles:
'~' is no good since it is already used in the tags of the event based calculator. The following characters cannot be used in Windows filenames (see https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247(v=vs.85).aspx):

```
< (less than)
> (greater than)
: (colon)
" (double quote)
/ (forward slash)
\ (backslash)
| (vertical bar or pipe)
? (question mark)
* (asterisk)
```
Not much is left. But we already have `_` (underscore) so I think we can just drop `|`.

g-weatherill:
OK, then we can stick with the underscore. Are spaces acceptable?

micheles:
Yes, spaces are acceptable.

PR #758, opened 2016-04-04
Accepted `.` and `|` as valid characters for source IDs

micheles:
Closes https://github.com/gem/oq-risklib/issues/753. While at it, I have also fixed an encoding issue when redirecting `oq-lite show all`, see https://ci.openquake.org/job/zdevel_oq-risklib/1437


openquake/commonlib/valid.py
```diff
@@ -184,20 +184,34 @@ def __call__(self, value):
...
+        if re.match(self.regex, value):
+            return value
+        raise ValueError(
+            'Invalid ID %r: the only accepted chars are a-zA-Z0-9_-' % value)
+
+MAX_ID_LENGTH = 100
+ASSET_ID_LENGTH = 100
+
+simple_id = SimpleId(MAX_ID_LENGTH)
+asset_id = SimpleId(ASSET_ID_LENGTH)
```
g-weatherill: Is it possible that the asset IDs might include the "." and "|" characters as the source IDs do?

g-weatherill:
LGTM

openquake/commonlib/valid.py
micheles: Right now no, the asset IDs continue to use the old regular expression `^[\w_\-]+$`

PR #764, opened 2016-04-05
Introduce an HDF5- serializable AssetCollection class

micheles:
Currently the assets are sent to the workers as Python objects. With this PR I introduce an AssetCollection class which is a thin wrapper over numpy arrays: that means that the data transfer and the memory occupation are *hugely* reduced. This is necessary for the California calculation, which involves 480,000+ assets. Just to give some figures, the data transfer goes down from 30.5 GB to 8.8 GB (over 3 times) and the memory occupation on the workers goes down by 12 times.

Tests and demos are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1446

PS: this PR breaks compatibility with Ubuntu 12.04 unless we backport the HDF5 libraries.

PPS: the `job_info` array has been replaced as an instance of `openquake.baselib.hdf5.LiteralAttrs` introduced in https://github.com/gem/oq-hazardlib/pull/434.


openquake/risklib/riskinput.py
```diff
@@ -37,117 +37,158 @@
...
+                 time_events=''):
+        self.cc = cost_calculator
+        self.time_event = time_event
+        self.time_events = time_events
+        self.array, self.taxonomies = self.build_asset_collection(
+            assets_by_site, time_event)
+        fields = self.array.dtype.names
+        self.loss_types = sorted(f for f in fields
+                                 if not f.startswith(FIELDS))
+        # hack: in Ubuntu 12.04 we cannot store empty lists while empty
```
ptormene: Please see if we want to remove the hack

ptormene:
A part from the comment, LGTM

PR #767, opened 2016-04-10
Fixed a bug with the CSV export of the ground motion fields

micheles:
The exporter was not being called due to a change of name. I have added an export test.
Everything is green: https://ci.openquake.org/job/zdevel_oq-risklib/1449/


raoanirudh:
LGTM

Issue #760, opened 2016-04-04
Reintroduce rupture tags in event loss tables?


raoanirudh:
The erstwhile `rup_tags` column in the loss tables for the `event_based_risk` calculator has now been replaced by the `rup_ids` column. However, the `rup_id`↔︎`rup_tag` map, which is present in the datastore under `etags`, is not available to the user to export.

The ideal solution would be to provide the `rup_tags` directly in the loss tables as before. An alternate solution would be to provide a csv exporter for `etags`.

PR #768, opened 2016-04-10
Changed the aggregate loss table exporter to export the event tags

micheles:
Fix https://github.com/gem/oq-risklib/issues/760.
Everything is green: https://ci.openquake.org/job/zdevel_oq-risklib/1450


raoanirudh:
I would suggest using the more descriptive name `event_tags` instead of `etags` for the first column. Other than that, LGTM.

PR #769, opened 2016-04-11
Added a check on the IMTs coming from the risk models

micheles:
Catalina had a computation with a wrong IMT in the vulnerability functions. With this PR the error is raised early, before running the computation and not in the middle of it.
The tests are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1451/


daniviga:
LGTM

PR #770, opened 2016-04-12
Saved the job.ini parameters in a datastore table

micheles:
Right now the parameters are stored as attributes of the datastore root, but it is nicer to store in their own table, as it was the case for the old engine. Notice that the companion https://github.com/gem/oq-hazardlib/pull/439/ is needed. The tests are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1455

Just run a calculation, open the datastore with the HDF5 viewer and look at the content of the table `oqparam`.


openquake/calculators/base.py
```diff
@@ -105,10 +105,8 @@ def save_params(self, **kw):
         """
         Update the current calculation parameters and save oqlite_version
         """
-        vars(self.oqparam).update(kw)
-        for name, val in self.oqparam.to_params():
-            self.datastore.attrs[name] = val
-        self.datastore.attrs['oqlite_version'] = repr(__version__)
+        vars(self.oqparam).update(oqlite_version=repr(__version__), **kw)
+        self.oqparam = self.oqparam
```
ptormene: This looks wrong

openquake/commonlib/oqvalidation.py
```diff
@@ -240,6 +256,7 @@ def set_risk_imtls(self, risk_models):
         for taxonomy, risk_functions in risk_models.items():
             for loss_type, rf in risk_functions.items():
                 imt = rf.imt
+                # imls = [round(iml, 3) for iml in rf.imls]
```
ptormene: Please remove this if not needed

ptormene:
LGTM

PR #772, opened 2016-04-12
Added a filtering on the minimum intensity

micheles:
During large risk calculations we generate a lot of ruptures which have a minimal impact on the losses, but a huge impact on the runtime: a substantial amount of the computation time is wasted computing losses generated by very small GMFs. In this pull request we introduce the concept of *minimum_intensity*:
GMFs with intensity smaller than the minimum are considered zero. Also, if a GMF is smaller than the minimum on all sites, the rupture generating it is completely discarded.
The user can enable this kind of rupture filtering by specifying a dictionary in the job.ini file,
for instance:

```
minimum_intensity = {'PGA': 0.01, 'SA(0,1)': 0.02, 'default': 0.05}
```

The default is used for the IMT not explicitly listed. It is also possible to write

```
minimum_intensity = 0.03
```

which is interpreted as

```
minimum_intensity = {'default': 0.03}
```

i.e. 0.03 is used as minimum intensity level for all IMTs.

If minimum_intensity is not specified, or the dictionary is empty, then the usual rupture filtered based
on the maximum distance is applied and a rupture is discarded only if it is far away from all sites.

Tests and demos are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1468/

This PR also adds a view `portfolio_loss` that can be used to compare the results with and without filtering; just run

```bash
$ oq-lite show portfolio_loss
```
to get the portfolio loss per each loss type and realization.


raoanirudh:
The error message provided when the user specifies the minimum intensity for some, but not all, of the IMTs used in their calculation could be improved. (Currently, it prints `KeyError: 'default'`)

It would be preferable to have the user specify the minimum intensities for each IMT explicitly. Thus, the `default` option could be avoided.

Apart from the above comments, LGTM.

__
Suggestions for future refinements of this filtering:

If a particular rupture is present multiple times in a SES, only one GMF could be generated for that rupture for each GSIM, by using the median values from the GSIM (setting `truncation_level = 0`) rather than by sampling from the distribution as being done now.

raoanirudh:
@micheles, please disregard the comment above regarding the error message. I see that a recent commit has already resolved this.

PR #773, opened 2016-04-13
oq-lite show all is now safe

micheles:
In the past the command `oq-lite show all` was automatically removing all unreadable datastores. This is now fixed because there is a very common case for unreadable datastore that must be considered valid, i.e. a change of version. The old datastore must not be deleted, since it can be read just going back to the version used to write it.


ptormene:
LGTM

PR #776, opened 2016-04-16
Moved the expected values for a couple of tests into text files

micheles:
This looks like an insignificant refactoring, but it pretty useful because the expected files can be regenerated automatically. On the other hand, if the expected values are stored in the Python code, then they must be regenerated manually when the algorithm change or the seeds change, i.e. very often these days. The tests are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1463


ptormene:
LGTM

PR #781, opened 2016-04-18
Changed the portfolio loss view

micheles:
I reduced the precision to 5 decimal digits and added a first column with the realization ordinal.

https://ci.openquake.org/job/zdevel_oq-risklib/1478


daniviga:
LGTM

Issue #784, opened 2016-04-19
Use different seeds for different realizations


micheles:
Currently the realizations in the case of a complex logic tree are not independent, so the ground motion fields generated by the same rupture for different realizations may be identical. The risk team suggested to solve this by giving different seeds to different realizations.

PR #785, opened 2016-04-19
Changes to the event based calculators

micheles:
This is a set of changes preliminary to the implementation of https://github.com/gem/oq-risklib/issues/784.

1. The ruptures are stored in a single sescollection dataset. Before, there was a sescollection for each tectonic model id.
2. In the case of sampling, I am not producing an event set per sample anymore. This is to make the case of full enumeration and sampling to use the same algorithm. When https://github.com/gem/oq-risklib/issues/784 will be implemented different realizations will produce different numbers even if the same path is sampled more than once.
3. The change in 2 solves the redundancy in the GMFs seen in the expected files in test_case_18, that
was introduced in https://github.com/gem/oq-risklib/commit/a3fb2492b8f32fce086d8f1e8735221274e43831 (see files `gmf-smltp_b1-gsimltp_AB-ltr_0.txt` and `gmf-smltp_b1-gsimltp_AB-ltr_1.txt` for instance). Also, it explains why we are now producing less ruptures in the file `case_17/expected/ses.xml`.
4. the `duration` is now consistently given by `oq.investigation_time * oq.ses_per_logic_tree_path` and not `oq.investigation_time * oq.ses_per_logic_tree_path * (oq.number_of_logic_tree_samples or 1)`
5. The rupture tags do not require the `col=` specification anymore and now they are ordered by the rupture serial number
6. The change of tag ordering explains why the numbers in a few tests are slightly different: the epsilons are associated differently to the ruptures
7. In `test_case_7` the parameter `ses_per_logic_tree_path = 10` has been increased to `ses_per_logic_tree_path = 200` so that the convergency to the classical hazard curves is better (at 20% in 10,000 years)

The tests are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1492


openquake/calculators/event_based.py
```diff
@@ -380,16 +388,15 @@ def build_eb_ruptures(
...
+        for ses_idx, num_occ in sorted(
                 num_occ_by_rup[rup].items()):
             for occ_no in range(1, num_occ + 1):
-                etag = 'col=%02d~ses=%04d~src=%s~rup=%d-%02d' % (
-                    col_idx, ses_idx, src.source_id, serial, occ_no)
-                etags.append(etag)
-        if etags:
-            yield EBRupture(rup, r_sites.indices, etags,
-                            src.trt_model_id, serial)
+                events.append((0, ses_idx, occ_no))
```
ptormene: Please add a comment about `0`

openquake/qa_tests_data/event_based_risk/occupants/expected/agg_curve.xml
```diff
@@ -12,13 +12,13 @@ xmlns:gml="http://www.opengis.net/gml"
     unit="people"
     >
         <poEs>
-            0.0384416 0.0231283 0.00508702 0.00239712 0.00139904 0.000499904 0.000399947 0.000199974 0.000199974 0.0
+            0.0384416 0.0240071 0.004888 0.00239712 0.00149888 0.000499904 0.000499904 0.000399947 0.000100017 0.0
```
ptormene: Please check

ptormene:
Besides minor comments, if changes in numbers are ok, LGTM

openquake/calculators/event_based.py
micheles: Done

openquake/qa_tests_data/event_based_risk/occupants/expected/agg_curve.xml
micheles: Even the occupants tests is indeed sensible to the epsilons, even if the report was not showing it. I have regenerated the report and checked manually that GMFs are the same and only the ordering of the epsilons is different.

PR #786, opened 2016-04-20
Added the information `require_epsilons` to the reports

micheles:
This is important for the risk calculators: we want to know from the report if the calculator required epsilons (i.e. there were vulnerability functions with nonzero coefficients of variation) or not.
The tests are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1489/

PS: I have removed from the report the information about the detailed information about the riskinputs, that was too detailed.


ptormene:
LGTM

PR #794, opened 2016-04-27
Reduced the runtime of tests case_3 and case_4 (event based risk)

micheles:
I removed some source slow sources and changed some parameters so that now the total runtime of the event based risk tests on my machine is down from 67 seconds to 38 seconds. See https://ci.openquake.org/job/zdevel_oq-risklib/1507


ptormene:
LGTM

PR #797, opened 2016-04-28
Worked around a limitation of numpy 1.6

micheles:
Setting trusty as default in Jenkins oq-risklib_zdevel was a bad idea. An issue with numpy 1.6 entered in master causing the breakage in https://ci.openquake.org/job/master_oq-risklib/2096/.
It is solved in this pull request. The default for oq-risklib_zdevel has been restored to precise.
We should raise the priority of the `/opt` project and make sure that we support numpy 1.8.2.

The tests are now green: https://ci.openquake.org/job/zdevel_oq-risklib/1512


micheles:
LGTM

PR #799, opened 2016-04-29
Added a reduced version of Miriam's island as a test

micheles:
This is fast now (~4 seconds) and it is a previously untested situation with a region.
The tests are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1524/


ptormene:
LGTM

Issue #426, opened 2015-10-11
Remove the HDF5 infinite loop warning


micheles:
Here is how you can produce it:

```bash
$ nosetests -vsx openquake/calculators/tests/event_based_risk_test.py:EventBasedRiskTestCase.test_case_2 openquake/calculators/tests/event_based_risk_test.py:EventBasedRiskTestCase.test_case_2bis
HDF5: infinite loop closing library
      D,G,A,S,T,F,FD,P,PL,FD,P,FD,P,E,E,SL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL,FL
```

Updating to the latest h5py does not solve the issue :-(

daniviga:
Is always reproducible? I remember a test that was giving the warning in 
- Ubuntu 14.04
- RHEL 7

but not in
- Ubuntu 12.04

micheles:
It is reproducible even on Windows. I confirm however that the above command raises no warning on Ubuntu 12.04.

PR #800, opened 2016-04-30
Fix the tests on Ubuntu 16.04

micheles:
Right now there is a segmentation fault in the tests due to the nastiness of the new HDF5 libraries. The segfault can be avoided by closing properly the datastore. This requires changing the test runner that needs to reopen the datastore to export the data and check the results. Also, due to numeric precision issues, the numbers are slightly different, so I have reduced the export precision to 5 digits after the decimal point. Even so, the numbers in event based risk case_master are different at the last digit: rather then reducing the precision further, I have skipped this test on Ubuntu 12.04. There is still a nasty error in some situation, when the datastore cannot be closed, but I am turned it into a warning. It only appears in the tests, not in real usage.

NB: this change may also solve the long standing issue https://github.com/gem/oq-risklib/issues/426

Here is a run for Ubuntu 12.04 and Ubuntu 14.04: https://ci.openquake.org/job/zdevel_oq-risklib/1527/


openquake/commonlib/export/risk.py
```diff
@@ -111,7 +111,7 @@ def export_avg_losses(ekey, dstore):
     avg_losses = dstore[ekey[0]].value
     rlzs = dstore['rlzs_assoc'].realizations
     assets = get_assets(dstore)
-    writer = writers.CsvWriter(fmt='%.6E')
```
ptormene: Please set the format as a constant to be used in all occurrences

micheles: Done

ptormene:
LGTM

PR #801, opened 2016-05-02
Refined the rupture weighting algorithm

micheles:
For each site affected by the rupture, we multiply by the number of assets on that site, so that ruptures affecting a lot of assets are heavier than ruptures affecting few assets. Unfortunately the task distribution does not improve much, so more refinements will be needed in the future.

The tests are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1525/


ptormene:
If tests are green, LGTM

PR #809, opened 2016-05-06
Fixed the test classical/case_18 and some cleanup

micheles:
The numbers change due to the bug fixed in  https://github.com/gem/oq-hazardlib/pull/448
Jenkins is happy: https://ci.openquake.org/job/zdevel_oq-risklib/1551/

PS: you will notice that the expected files for the disaggregation test disagg/case_1 have changed: the reason is that I have set the parameter `concurrent_tasks to 4`. That test is sensitive to that parameter, probably due to numerical rounding errors. Without setting that parameter the test could pass on a laptop and not on Wilson which has more cores, nor on Jenkins. It is likely that the numbers are also sensitive to the ordering of the tasks. We should think more on the disaggregation calculator in the future.


g-weatherill:
LGTM

PR #810, opened 2016-05-09
Added a view displaying the calculation times by source typology

micheles:
The engine already displays the calculation times for each source in a table called source_info. Here I am adding a view over that table displaying the total times depending on the source typology. So one can sees if the calculation is dominated by PointSources or FaultSources or ...
This view is included in the full report.
The tests are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1550/
The generated reports are visible here: http://artifacts.openquake.org/demos/oq-engine/zdevel/precise/1801/jobs-2016-05-09.html#tabs-4


ptormene:
LGTM

PR #816, opened 2016-05-12
Inferring the minimum_intensity from the vulnerability functions

micheles:
As  requested by Cata and Anirudh, see https://ci.openquake.org/job/zdevel_oq-risklib/1572.

There is no change in the numbers, but a big change in terms of performance and memory occupation. Now the filtering is done in the hazard side, before passing the GMFs to the risk. For extra safety the filtering in the risk side is still there.


raoanirudh:
LGTM

PR #818, opened 2016-05-14
Standardized the risk_writers to use exponential notation with 5 decimal digits

micheles:
Now also the risk_writers use the same format and precision as the other exporters. Notice that I have removed the obsolete risk writers tests: they are now redundant and hard to maintain.
Nothing was broken: https://ci.openquake.org/job/zdevel_oq-risklib/1577


ptormene:
LGTM

PR #821, opened 2016-05-16
Reduced the number of tasks generated in build_riskinputs

micheles:
We discovered with the Bay area classical_risk calculation that we were generating `I` times more tasks than needed, where `I=5` was the number of IMTs. This PR fixes the issue. The tests are running here https://ci.openquake.org/job/zdevel_oq-risklib/1584

PS: while at it, I am also measuring the time and memory spent to build the risk inputs in the event based risk calculator.


raoanirudh:
LGTM

PR #822, opened 2016-05-17
UCERFEventBasedRuptureCalculator

micheles:
This rupture calculator converts the UCERF ruptures stored in HDF5 format into EBRuptures suitable for use in the engine. Then the GMFs are computed. The tests are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1596


openquake/calculators/ucerf_event_based.py
```diff
@@ -0,0 +1,763 @@
...
+        self.hdd = hdd
+        self.aspect = aspect
+        self.usd = upper_seismogenic_depth
+        self.lsd = lower_seismogenic_depth
+        self.msr = msr
+        self.mesh_spacing = mesh_spacing
+        self.tectonic_region_type = trt
+        self.seed = random.randint(0, MAX_INT)
+        self.rnd = None
+        self.integration_distance = 1000
```
micheles: Why this is hard-coded to 1000? I guess it should be passed from outside.

openquake/calculators/ucerf_event_based.py
```diff
@@ -0,0 +1,763 @@
...
+        self.sites = None
```
micheles: Should the sites be passed from outside?

micheles:
Let's merge this since it does not break anything and let's continue the work in the future.

PR #825, opened 2016-05-23
Fast filtering

micheles:
Companion of https://github.com/gem/oq-hazardlib/pull/455. The tests are green: https://ci.openquake.org/job/zdevel_oq-risklib/1619/

I run a calculation for South America which was dominated by the filtering. The difference in time is from   4,307 seconds to 245 seconds, i.e. a 18x speedup!


ptormene:
LGTM

PR #826, opened 2016-05-24
Added a test calculation crossing the International Date Line

micheles:
This was missing and it is needed to test properly https://github.com/gem/oq-risklib/pull/825.
The input files were provided by Graeme and reduced so that the computation runs in few seconds,
see https://ci.openquake.org/job/zdevel_oq-risklib/1607


g-weatherill:
LGTM

Issue #740, opened 2016-03-14
hazard_maps =false and uniform_hazard_spectra = true makes no sense


micheles:
This happened to Yen-Shin: probably we should generate the hazard_maps if uniform_hazard_spectra is True in any case (perhaps with a warning).

g-weatherill:
For a single site or a small number of sites it actually makes perfect sense that the user wishes to view the uniform hazard spectra but has no interest in the maps. Ultimately the internal calculation to get the IMLs for the required PoE is the same but there is no reason to store or export the maps in this case.

micheles:
Technically right now the UHS can be computed only if the hazard maps are available in the datastore. Users  keep getting confused (see https://groups.google.com/d/msg/openquake-users/zCW7gypguiQ/ApceZbeEBgAJ). The solution I see is to compute the hazard maps if  uniform_hazard_spectra is True if hazard_maps is not specified.

PR #828, opened 2016-05-24
Make uniform_hazard_spectra=True and hazard_maps=False valid

micheles:
Closes https://github.com/gem/oq-risklib/issues/740. If  uniform_hazard_spectra is true the hazard maps are generated even if hazard_maps is false, however they are not exported. This is a compromise, a clean solution would be to move the generation of hazard maps and uhs curves in a post-processing calculator able to compute the maps and the spectra independently. The tests are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1614/


g-weatherill:
LGTM

PR #824, opened 2016-05-19
Print the size of the RlzsAssoc instance

micheles:
Here is an example of usage:

```bash
$ oq-lite info demos/hazard/LogicTreeCase2ClassicalPSHA/job.ini 
<RlzsAssoc(size=324, rlzs=324)
...>
=============== =======
attribute       nbytes 
=============== =======
csm_info        178,320
rlzs_assoc      60,671 
rlzs_by_smodel  53,426 
gsim_by_trt     16,778 
gsims_by_trt_id 8,928  
=============== =======
```
It is clear that we could reduce the data transfer by removing the csm_info attribute. This is left for future work. The tests are green: https://ci.openquake.org/job/zdevel_oq-risklib/159

PR #829, opened 2016-05-25
Reduced the data transfer of the realization association object

micheles:
This is the dominating factor in computations with few sites, few sources and a big logic tree, such as our demo LogicTreeCase2ClassicalPSHA. By removing the csm_info object from the RlzsAssoc object the data transfer is reduced by 3 times, from  45,348,336 to 14,926,032 bytes, as it was discovered in https://github.com/gem/oq-risklib/pull/824.
The tests are running here: https://ci.openquake.org/job/zdevel_oq-risklib/1617


openquake/calculators/ucerf_event_based.py
```diff
@@ -741,12 +742,25 @@ def execute(self):
...
         return ruptures_by_trt_id
 
+    def agg(self, acc, val):
+        """
+        Aggregated the ruptures and the calculation times
+        """
+        for trt_id in val:
+            ltbrid, dt = val.calc_times[trt_id]
+            info = source.SourceInfo(
+                trt_id, ltbrid, 'UcerfBranch', 1, 0, 0, 0, dt)
```
ptormene: Please use the names of parameters, to make it more readable

ptormene:
LGTM

